PR #4, opened 2016-05-10
[MRG] Better connection plots

### LINKED ISSUE #3: Use images / adaptive point size for large synapses?
At the moment, if you ``brian_plot(S)`` for a large ``Synapses`` you just get a big blue square and you can't see any structure in it. Perhaps we could either have an image rather than scatter plot, or adaptively decide the size of the scatter symbols based on the number of synapses?

mstimberg:
Following our discussion in #3, I added a new `plot_type` argument to `plot_synapses` which can be `image`, `scatter`, or `hexbin`, creating the respective plot. `plot_synapses` will default to `scatter` (which I think is the most universally applicable), but `brian_plot` will chose a plot type based on the following heuristic:
* `image` if there are less than 1000 source and less than 1000 target neurons
* otherwise: `scatter` if there are less than 10000 synapses in total
* otherwise: `hexbin`

I think this should give something somewhat useful in all cases , we can always tweak the details in the future. If that's fine with you, then I'll start preparing the packaging so that we can do the glorious 0.1 release :-)

Closes #3


thesamovar:
Yep, go for it!

PR #5, opened 2016-05-12
[MRG] Prepare 0.1 release

mstimberg:
Ok, this should be all to do a 0.1 release, let me know if you think there are still some changes necessary. I'd like to to use `brian2tools` also a bit to prototype release/documentation/packaging stuff that we might want to use in the long term in Brian 2 as well -- relying as little as possible on custom-written code and use mostly existing solutions out there. I'm much less worried if we break something for `brian2tools` than for `brian2` :)

For now, here are a few differences that I am testing out:
* the documentation does not rely on any custom `briandoc` extension but uses `napoleon` and `sphinxdoc-api` to auto-generate the reference. This works fine so far, but has one disadvantage: you cannot shortcut references, i.e. you cannot write `` `brian_plot` `` but you'll have to write `` `~brian2tools.plotting.base.brian_plot` ``. I think we agree that this is annoying... I did not investigate this in any detail yet, maybe there's an easy solution to make this work with `napoleon`
* the version number is stored in a single place and automatically set based on the git tag (using the "official" `setuptools_scm` package). One nice feature is that this will automatically set the version number to something meaningful when you run it directly from the git repository (so we can ask users for their version and get a precise response). Let's see whether this works well in practice, though.
* in the long run, I'd like to do the upload to PyPI automatically from the test servers, in the same way we already do for conda. For now, this is still an explicit step.


thesamovar:
No changes necessary I think, this all looks great to me.

mstimberg:
After struggling a bit with the documentation build, all the tests pass now, I'll therefore go ahead and merge. Let's see whether the release scripts actually work :-)

PR #14, opened 2016-10-06
[MRG] Plot synapses fixes

mstimberg:
Fixes two bugs in the `plot_synapses` function.

### DIFF
--- a/brian2tools/plotting/synapses.py
+++ b/brian2tools/plotting/synapses.py
@@ -182,7 +182,10 @@ def plot_synapses(sources, targets, values=None, var_unit=None,
             origin = kwds.pop('origin', 'lower')
             interpolation = kwds.pop('interpolation', 'nearest')
             axes.imshow(matrix, origin=origin, interpolation=interpolation,
-                        cmap=cmap, norm=norm, **kwds)
+                        cmap=cmap, norm=norm,
+                        extent=(min(unique_sources), max(unique_sources),
+                                min(unique_targets), max(unique_targets)),
+                        **kwds)
 
         # Add the colorbar
         locatable_axes = make_axes_locatable(axes)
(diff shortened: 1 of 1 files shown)


brian2tools/plotting/synapses.py
```diff
@@ -182,7 +182,10 @@ def plot_synapses(sources, targets, values=None, var_unit=None,
             origin = kwds.pop('origin', 'lower')
             interpolation = kwds.pop('interpolation', 'nearest')
             axes.imshow(matrix, origin=origin, interpolation=interpolation,
-                        cmap=cmap, norm=norm, **kwds)
+                        cmap=cmap, norm=norm,
+                        extent=(min(unique_sources), max(unique_sources),
+                                min(unique_targets), max(unique_targets)),
+                        **kwds)
```
thesamovar: I think if you want the pixels to be centred on the correct values, you need to do something like subtract 0.5 from the min values and add 0.5 to the max values, otherwise the centres won't be consistent.

brian2tools/plotting/synapses.py
```diff
@@ -207,7 +210,10 @@ def plot_synapses(sources, targets, values=None, var_unit=None,
             vmin = kwds.pop('vmin', 1 if values is None else None)
             plotted = axes.imshow(matrix, origin=origin,
                                   interpolation=interpolation,
-                                  vmin=vmin, **kwds)
+                                  vmin=vmin,
+                                  extent=(min(sources), max(sources),
+                                          min(targets), max(targets)),
+                                  **kwds)
```
thesamovar: Same here.

mstimberg:
Good point about the centering -- it was not something introduced by this change, though, i.e. it was incorrect before as well. I fixed it and as a bonus I forced the axis ticks to be integers -- it looked odd to have a "source index" 88.5 appearing when you zoomed in...

thesamovar:
Nice!

PR #17, opened 2018-05-23
[WIP] nmlimport package: importing morphologies from nml file

kapilkd13:
This PR features current progress in implementing nmlimport package that allows importing morphologies from nml file.

### DIFF
--- a/brian2tools/tests/samples/sample1.cell.nml
+++ b/brian2tools/tests/samples/sample1.cell.nml
@@ -55,7 +55,7 @@
             </segment>
 
             <segment id="6" name="basal0">
-                <parent segment="0" fractionAlong="1.0"/>
+                <parent segment="0" fractionAlong="0.0"/>
                 <proximal x="0.0" y="17.0" z="0.0" diameter="4.0"/>
                 <distal x="0.0" y="-50.0" z="0.0" diameter="4.0"/>
             </segment>


kapilkd13:
Hi @mstimberg I am writing test for `load_morphology` function. Here I am comparing the final morphology object attributes to pre determined values to assert our function is correct. I am checking attributes like length,distance, coordinates and area. Although other three are providing correct result, area is failing assertion. see line https://github.com/brian-team/brian2tools/pull/17/files#diff-1a36c24e3b7044d2e6d50df875dc2968R63
output that we get:
```
=================================== FAILURES ===================================
_____________________________ test_load_morphology _____________________________

    def test_load_morphology():
        morphology = load_morphology("samples/sample1.cell.nml")
        assert_equal(morphology.distance,[ 8.5]*um)
        assert_equal(morphology.length,[ 17.]*um)
        assert_equal(morphology.coordinates,[[0.,0.,0.],[0.,17.,0.]]*um)
    
>       assert_equal(morphology.area,[1228.36272755]*um2)
E       AssertionError: 
E       Arrays are not equal
E       
E       (mismatch 100.0%)
E        x: Quantity([  1.228363e-09])
E        y: Quantity([  1.228363e-09])
```
Can you think on any reason? Can float-point precision be behind it.

mstimberg:
> Can you think on any reason? Can float-point precision be behind it.

Yes, it's very like *almost* the same value. For these kind of comparisons, rather use `assert_allclose` instead of `assert_equal`.

kapilkd13:
Hi @mstimberg  I have added tests and docstring. Please have a look. I haven't updated sphinx doc to auto import docstring as due to some reason sphinx is not working on my pc for this particular project.
`make html` gives me 
```
sphinx-build: error: unrecognized arguments: /home/kapil/github/brian2tools/brian2tools /home/kapil/github/brian2tools/brian2tools/tests
```
I suspect it has to do something with the fact that I have `pip develop` this module instead of building it.

Also, passing `file object` works with `libNeuroML` (it is using lxml), although in docs it mention for file location only. I have updated `load_morphology` accordingly.

mstimberg:
> due to some reason sphinx is not working on my pc for this particular project.

Mmm, that's odd, it should work with `pip install -e .` (that's what I'm using) or `python setup.py develop`. What happens if you directly run `sphinx-build -b html -d ../docs/doctrees   . ../docs/html` in `docs_sphinx`? That should be what the makefile does.

> I haven't updated sphinx doc to auto import docstring

You don't need to do anything to import the docstrings. The `conf.py` file automatically runs `sphinx-apidoc` to create the reference documentation for all files in the package.

Could you convert your docstrings to the [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) format? I find it much more readable in the source code and it is what we are using in the rest of `brian2` and `brian2tools`.

kapilkd13:
Hi @mstimberg  I have updated the docstring. I tried the sphinx build command but it also throws same error. Travis build is failing right now because we need `libneuroml` for testing but its latest package is not available for conda. Since tests are a part of conda recipe, pip installing libneuroml in travis file after the conda environment is built doesn't work either. We can create and upload `libneuroml` package for conda and use that. Is there some other way? Or we can move tests out of conda recipe and into travis file, would that work?

kapilkd13:
@mstimberg  I have made  conda recipe for libNeuroML, can you have a look and see if it is good to go.
https://github.com/kapilkd13/staged-recipes/blob/libneuroml/recipes/libneuroml/meta.yaml

mstimberg:
Hi @kapilkd13 , this is looking very good! Just some minor things to make it perfect:
* Since libneuroml can be used with Python 2 and 3 without changes and is "pure Python", it can be a [`noarch` package](https://conda-forge.org/docs/meta.html#building-noarch-packages) (but we should check back with Padraig, setup.py only mentions Python 2 in its classifiers)
* The installation should [use `pip`](https://conda-forge.org/docs/meta.html#use-pip) which therefore needs to be a build dependency
* In addition to the `dev_url`, you could also add http://libneuroml.readthedocs.io as the `doc_url`

kapilkd13:
@mstimberg I forgot to mention I have developed the test script we were talking about. Its in the last commit here

mstimberg:
No worries, I saw the commit :)

kapilkd13:
@mstimberg  I added function to generate proximal distance from parent and its working fine. But their is another issue, while validating we are checking if the proximal distance of a segment is matching to the distal of its parent but in some files their is `fraction along` is used and in these cases proximal point do not overlap with distal. This throws validation error. In this file https://github.com/NeuroML/NeuroML2/blob/master/examples/NML2_SimpleMorphology.nml
we are getting
```
brian2tools.nmlimport.nml.ValidationException: <Segment|3|Spine> and its parent segment <Segment|2|MainDendrite2> are not connected!!
```

mstimberg:
Hi, this is fine, we don't support these morphologies yet, anyway. I think the best thing would be to raise a `NotImplementedError` if a segment defines `fractionAlong` with a value different from 1.

kapilkd13:
Hi @mstimberg I added runtime dependencies to conda recipe. Should we push it now https://github.com/kapilkd13/staged-recipes/blob/libneuroml/recipes/libneuroml/meta.yaml

mstimberg:
> Hi @mstimberg I added runtime dependencies to conda recipe. Should we push it now https://github.com/kapilkd13/staged-recipes/blob/libneuroml/recipes/libneuroml/meta.yaml

It is looking good, but some of the tests fail for me when building/installing/testing the package locally. It seems to be an issue with the `jsonpickle` library, if I install the 0.7.2. version from pip (unfortunately it's not packaged for conda-forge), the test runs fine. Could you maybe have a quick look at the bug, maybe it's something that is easily fixable in `libneuroml` (possibly via a `try...except` that tries something that works for the latest versions and falls back to something  that works for earlier versions of `jsonpickle`)?

Oh, and I think you can remove vellamike from the maintainers, according to Padraig he left academia so I guess he's not particularly interested in maintaining the package.

brian2tools/nmlimport/nml.py
```diff
@@ -31,7 +31,7 @@ def validate_morphology(segments):
         for segment in segments:
 
             if segment.parent is not None:
-                if segment.parent.fraction_along != 1:
+                if segment.parent.fraction_along not in [0,1]:
```
mstimberg: I don't think we correctly support `fraction_along=0`, it would mean to connect to the proximal end of a segment.

kapilkd13: Currently I was checking if it is connected proximally [here](https://github.com/brian-team/brian2tools/pull/17/files#diff-b80b7aa670e817379b0298762aa269d5R21) I guess we should remove it then :)

kapilkd13:
Hi @mstimberg how are you testing conda recipe locally, currently I am using `conda-build .` to build recipe after adding 
```
  commands:
    -  py.test --pyargs neuroml
  requires:
    - pytest
``` 
to `meta.yml` file. On running I am getting errors like
```
  def read_neuroml2_file(nml2_file_name, include_includes=False, verbose=False,
                           already_included=[], print_method=print_, optimized=False):
    
        print_method("Loading NeuroML2 file: %s" % nml2_file_name, verbose)
    
        if not os.path.isfile(nml2_file_name):
            print_method("Unable to find file: %s!" % nml2_file_name, True)
>           sys.exit()
E           SystemExit

../_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehol/lib/python3.6/site-packages/neuroml/loaders.py:235: SystemExit
```
Is this the right approach? Also how did you pip installed specific version of jsonpickle with conda recipe. is pip allowed in meta file?

mstimberg:
Sorry, I should have been clearer about this. I don't do any change to the `meta.yaml` file (I don't think that you can run all the tests during the package build, since the necessary test files are not part of the package), instead I build it as it is. I have created a new conda environment for testing the package, and then I install it with:
```command
conda install -c conda-forge --use-local libneuroml
```
Because of the `--use-local`, it will use the package that I've previously built. I then run the tests in `neuroml/test` directory of the source repository (with nosetests, but pytest it should work as well). As you said, you cannot install from pip in `meta.yaml` -- I simply removed the conda package for `jsonpickle` in my test environment and installed the older version with `pip install jsonpickle==0.7.2`.

pgleeson:
@kapilkd13 I agree that the case fraction_along != 0 or 1 could be ignored, as it would complicate the conversion of the model to Brian & not produce correct electrical connectivity, but the case fraction_along = 0 should be fine: it explicitly states that the new segment is connected to the proximal point of the parent, and that should be fine to represent in Brian format.

mstimberg:
@pgleeson I agree that Brian does support it in principle. The reason why I told @kapilkd13 to ignore it for now is that the current approach for converting the NeuroML morphology to Brian is to first convert it into a point-based format like the one used by SWC, so that we can then use Brian's existing import facility for this representation. With that SWC-like format, we only have a simple "parent" link, so we cannot express the difference between a proximal and distal connection.

kapilkd13:
Hi @mstimberg  I tried few things to resolve `jsonpickle` issue but wasn't very successful. libneuroml is using it for a very trivial task of loading json and it shouldn't throw this error. I do found an active issue on jsonpickle repo https://github.com/jsonpickle/jsonpickle/issues/192 related to this. I guess its some issue with jsonpickle itself..

mstimberg:
Hi @kapilkd13 , I looked into this in quite a bit of detail and added my comments to the github issue you linked. The specific issue seems to be a problem in `jsonpickle` itself, but there's also an option that we can enable in newer `jsonpickle` versions that does the trick: NeuralEnsemble/libNeuroML/pull/72

Note that during testing, I had to clear the database, otherwise it would not overwrite the incorrect JSON with the fixed version:
```pycon
>>> import pymongo
>>> client = pymongo.MongoClient('localhost', 27017)
>>> client.drop_database('mike_test_db3')
>>> client.drop_database('test_db_4')
```

mstimberg:
> I also want to ask if I should push segment group related commits on this branch only or should we get this merge and start our work on new branch.

I'd say: until we have a `libneuroml` package on conda-forge (and therefore can run the automatic test suite in its current form) you can add your changes here. As soon as this is the case, we can finish this branch and merge it, and then start to work on new features in separate pull requests.

mstimberg:
Can you have another go at the conda-forge recipe with the updated version of `libneuroml` that Padraig just uploaded to pypi (https://pypi.org/project/libNeuroML/)? It should now hopefully work with a recent version of `jsonpickle`. If it does, I think it is ready to for a pull request in conda-forge (you could maybe squash all commits into one before proposing it, though).

kapilkd13:
@mstimberg  I tried with new libneuroml version but I am still getting the error. I modified `meta.yaml` file, updated libneuroml repo for any new changes, dropped mongo table. Then I created a new conda env for testing but I am still getting these errors. Can you test it on your system and verify.

kapilkd13:
@mstimberg  I have created a new branch for conda recipe. Here is a link to final recipe https://github.com/kapilkd13/staged-recipes/blob/add_libneuroml/recipes/libneuroml/meta.yaml
Everything is working fine now. Actually I forgot to switch to development branch after pulling changes.

mstimberg:
Great, I can confirm that it works fine on my machine as well. Please go ahead and open the pull request for inclusion in conda-forge.

kapilkd13:
Hi @mstimberg  I was working on morphology function. I have a doubt, while reading a nml file how should we determine if the given set of segments belong to a section. What I am thinking is to create a tree of morphology object and perform DFS and consider link of segments that have only one children (ex. A--B--C--D) as one section. Should we do this for intermediate nodes also. I mean
A-----B-----C---D---E----G
         |              |
        C             F
In this tree with soma at A, we can consider E---G as one section, but should we also consider C--D as one section as C has only one child, giving us
A-----B-------sec1---sec2
         |              |
        C             F

kapilkd13:
Hi @mstimberg  I have pushed current development, moving towards class implementation and changing morphology generation method. Its pretty rough right now but it will be the basic structure for us. please have a look.
Command like this should work
```
from brian2tools.nmlimport.nml import *
s=MorphologyTree('nml-file')
plot_morphology(s.morphology_obj)
```

mstimberg:
Hi @kapilkd13, the approach looks good at the first sight, but I did not go through it in much detail. To make things a bit easier to navigate it would be good to consistently name internal helper functions with starting underscores, to make clear that they are not supposed to be called by the user (e.g. `get_morphology_dict`).

kapilkd13:
Hi @mstimberg I am having some issues when trying to use `unknown` for section names.
suppose  we have
soma----some-section------apical1----apical2
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;|
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;apical3
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;|
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;apical4

here apical1 and 2 forms one section and apical3 and 4 form another. using unknown for both section gives error as both section will be connected to parent section `some-section` with same id `unknown`. This problem also occurs in case of name_heuristic as usually we would remove integer from end of the name of section which will again result in two section with same name `apical`. 
Currently what I am doing is, if we encounter a section with multiple segments in it then `section_name== name of the first segment in the section` i.e in above case it would be apical1 and apical3.

mstimberg:
Good point! I think the best solution would be:
* without heuristic ("unknown"): number the sections yourself. I would not actually use the word "unknown", but maybe something more neutral like "sec" (for "section"). E.g. in your above example, you'd name the "apical1 - apical2" section "sec1" and the "apical3 - apical4" section "sec2" (or you use "sec", "sec2", not sure...)
* with heuristic, you could try to give a combined name, e.g. call the first one "apical1_2" and the second one "apical3_4" -- just using the first name is a good short-term solution, though.

mstimberg:
> Is there a way to include data files in conda's test environment? Would including these files as package_data inside setup.py file work?

Yes, that should work. We'll actually have to do this if we want to make the testing work for users that install via `pip` -- the other options would only work within the test environment.

kapilkd13:
Hi @mstimberg I have uploaded sphinx doc for users and the tests are also passing now :) Can you check and suggest improvements in sphinx docs. 
P.S I still can't get sphinx docs to build on my local system when calling from brian2tools folder, so I copied all the docs to some other project and tested there and later copied the docs back to brian2tools. Can you build it on your system ? Also is there a way to check docs that are built as part of travis sphinx doc test.

mstimberg:
I think your branch does not contain the changes we did to fix the sphinx issues. After merging in the changes from brian-team's master branch, it should build fine.
Is there anything else you want to do for this PR before I review everything (not only the docs)?

kapilkd13:
ah, I forgot to rebase my branch against master. Docs are building now, Thanks :) Please go ahead and review this PR. In the meantime I will look at the biophysical properties extracted from the file.

brian2tools/nmlimport/nml.py
```diff
@@ -0,0 +1,349 @@
...
+    except ValidationException as v:
+        logger.error("Validation error: {0}".format(v))
+        raise
+
+    except Exception as e:
+        logger.error("Exception occured: {0}".format(e))
+        raise
+
+
+class NmlMorphology(object):
```
mstimberg: For consistency with the NeuroML exporter, this should probably be called `NMLMorphology`

brian2tools/nmlimport/nml.py
```diff
@@ -0,0 +1,349 @@
...
+
+        resolved_ids = {}
+        for group in m.segment_groups:
+            grp_ids = self.get_segment_group_ids(group.id, m)
+            id_map = get_id_mappings(m.segments)
+            resolved_ids[group.id] = list(set([id_map[grp_id] for grp_id in
+                                               grp_ids]))
+        return resolved_ids
+
+    # Helper function to read .nml file and return document object
```
mstimberg: Even though docstrings for helper functions are not mandatory, I'd put this comment as a docstring into the function (which makes it accessible to the help, etc.)

brian2tools/nmlimport/nml.py
```diff
@@ -0,0 +1,349 @@
...
+            # Validating NeuroML file
+        validate_neuroml2(deepcopy(file_obj))
+        logger.info("Validated provided .nml file")
+
+        # Load nml file
+        doc = loaders.NeuroMLLoader.load(file_obj)
+        logger.info("Loaded morphology")
+        return doc
+
+    '''
```
mstimberg: The docstring here should go into the function as well.

brian2tools/nmlimport/nml.py
```diff
@@ -0,0 +1,349 @@
...
+    Helper function that determines if the given segment belongs to the 
+    passed section or not as per our heuristic.
+    '''
+    def _is_heurestically_sep(self, section, seg_id):
+        root_name = section.name.rstrip('0123456789_')
+        seg = self.seg_dict[self.children[seg_id][0]]
+        return not seg.name.startswith(root_name)
+
+    '''
```
mstimberg: Same here

brian2tools/nmlimport/nml.py
```diff
@@ -0,0 +1,349 @@
...
+    Helper function that generate the new section name based on whether 
+    name_heuristic is set to True or not.
+    '''
+    def _get_section_name(self, seg_id):
+        if not self.name_heuristic:
+            self.incremental_id = self.incremental_id + 1
+            return "sec" + str(self.incremental_id)
+        return self.seg_dict[seg_id].name
+
+    '''
```
mstimberg: Same here

brian2tools/nmlimport/nml.py
```diff
@@ -0,0 +1,349 @@
...
+                    m = re.search(r'\d+$', seg_name)
+                    section.name = '{}_{}'.format(section.name, m.group())
+                    self._create_tree(section,
+                                      self.seg_dict[self.children[seg.id][0]])
+            else:
+                self._create_tree(section,
+                                  self.seg_dict[self.children[seg.id][0]])
+        return section
+
+    '''
```
mstimberg: Same here

docs_sphinx/user/nmlimport.rst
```diff
@@ -0,0 +1,322 @@
+NeuroML importer (nmlimport library)
+================
```
mstimberg: The underline should be the full length of the title (but actually not sure that we need the "nmlimport library" part her).

docs_sphinx/user/nmlimport.rst
```diff
@@ -0,0 +1,322 @@
...
+(`NML project website <https://neuroml.org/>`_).
+
+.. contents::
+    Overview
+    :local:
+
+Working example
+---------------
+
+As a demonstration, we will use nmlimport library to generate morphology and
```
mstimberg: we will use *the* nmlimport package to generate *a* morphology...

docs_sphinx/user/nmlimport.rst
```diff
@@ -0,0 +1,322 @@
...
+extract other related information from the ``pyramidal cell's`` nml file. You
```
mstimberg: Putting "pyramidal cell's" in verbatim looks a bit odd here -- it would make sense if you used the actual file name, though (`pyr_4_sym.cell.nl`).

docs_sphinx/user/nmlimport.rst
```diff
@@ -0,0 +1,322 @@
...
+can download it from `here <https://github
+.com/OpenSourceBrain/ACnet2/blob/master/neuroConstruct/generatedNeuroML2/pyr_4_sym.cell.nml>`_.
+
+
+.. code:: python
+
+    from brian2tools.nmlimport.nml import NmlMorphology
+    nml_object = NmlMorphology('pyramidal.cell.nml',name_heuristic=True)
```
mstimberg: Would be good to use the file name `pyr_4_sym.cell.nml` here -- otherwise users have to rename it first. Also, there's a space after the comma missing :smirk:

docs_sphinx/user/nmlimport.rst
```diff
@@ -0,0 +1,322 @@
...
+
+This call provides us the ``nml_object`` that contains all the information
+extracted from `.nml` file. When ``name_heuristic`` param is set to True
```
mstimberg: This sentence is a bit long and difficult to understand. Maybe divide it in two?

mstimberg: (with "this sentence" I mean: "When ``name_heuristic`` param is set ..."

docs_sphinx/user/nmlimport.rst
```diff
@@ -0,0 +1,322 @@
...
+name will be created by combining names of the inner segments of the section
+else if set to False, all linearly connected segments combines and form a
+section and naming convention `sec{unique integer}` is followed.
+
+|
+- To obtain morphology object.
+
+.. code:: python
+
+    morphology=nml_object.morphology_obj
```
mstimberg: Spaces around `=` :smile:

docs_sphinx/user/nmlimport.rst
```diff
@@ -0,0 +1,322 @@
...
+|
+
+- To obtain topology information.
```
mstimberg: Maybe this section could be shortened a bit. How about saying: "With this `Morphology` object, you can use all of Brian's functions to get information about the cell:" and then having a long code block (with these interactive `>>>` prompts). I don't think you need to say that `morphology.area` gives you the area, etc.

docs_sphinx/user/nmlimport.rst
```diff
@@ -0,0 +1,322 @@
...
+    plot_morphology(morphology)
+
+.. figure:: ../images/pyramidal_morphology.png
+   :scale: 80 %
+   :alt: Pyramidal cell's morphology plot.
+
+   Pyramidal cell's Morphology plot.
+
+
+- To get the resolved group ids of an nml `SegmentGroup`. This returns a
```
mstimberg: As discussed, I think it would be good to first show here a shortened version of the NeuroML file. I'd also add a little explanation of what a `SegmentGroup` is. BTW: You should use two backticks, i.e. ` ``SegmentGroup``  ` because `SegmentGroup` is not an object defined anywhere in the documentation.

docs_sphinx/user/nmlimport.rst
```diff
@@ -0,0 +1,322 @@
...
+    # Output
+    {'soma': [0], 'apical0': [1], 'apical2': [2], 'apical3': [3], 'apical4':
+    [4], 'apical1': [5], 'basal0': [6], 'basal1': [7], 'basal2': [8], 'all':
+    [0, 1, 2, 3, 4, 5, 6, 7, 8], 'soma_group': [0], 'dendrite_group':
+    [1, 2, 3, 4, 5, 6, 7, 8], 'apical_dends': [1, 2, 3, 4, 5],
+    'middle_apical_dendrite': [3], 'thalamic_input': [5], 'basal_dends':
+    [8, 6, 7], 'basal_gaba_input': [6], 'background_input': [7]}
+|
+
+- To get the information of all the child segments of a parent segment. This
```
mstimberg: Not sure we need this, you can also use the children information stored in Brian's morphology

docs_sphinx/user/nmlimport.rst
```diff
@@ -0,0 +1,322 @@
...
+                <include segmentGroup="basal0"/>
+            </segmentGroup>
+
+            <segmentGroup id="background_input">
+                <include segmentGroup="basal1"/>
+            </segmentGroup>
+        </morphology>
+    </cell>
+
+Handling FractionAlong value
```
mstimberg: I would rather explain this in the section title (e.g. "Handling sections not connected at the distal end")

docs_sphinx/user/nmlimport.rst
```diff
@@ -0,0 +1,322 @@
...
+--------------------------------
```
mstimberg: Underline needs to have the same length as the title :smile:

kapilkd13:
Hi @mstimberg , I have made the suggested changes. But due to some reason travis is failing for python2. It looks like there is some issue with `conda build` in python2. Please have a look at https://travis-ci.org/brian-team/brian2tools/jobs/402402431

mstimberg:
> But due to some reason travis is failing for python2. It looks like there is some issue with conda build in python2.

There was a problem in the `conda` package itself, but it got updated a few hours ago. I restarted the tests and they now pass.

brian2tools/nmlimport/nml.py
```diff
@@ -0,0 +1,417 @@
...
+        validate_neuroml2(deepcopy(file_obj))
+        logger.info("Validated provided .nml file")
+
+        # Load nml file
+        doc = loaders.NeuroMLLoader.load(file_obj)
+        logger.info("Loaded morphology")
+        return doc
+
+
+    def _is_heurestically_sep(self, section, seg_id):
```
mstimberg: "heurestically" --> "heuristically"

brian2tools/nmlimport/nml.py
```diff
@@ -0,0 +1,417 @@
...
+        section: SectionObject
+            Object of class SectionObject that contains information about
+            segments belonging to same section.
+        seg_id: int
+            segment id of given segment
+
+        Returns
+        -------
+        bool
+            Returns true if the given segment is not heurestically connected
```
mstimberg: same here

brian2tools/nmlimport/nml.py
```diff
@@ -0,0 +1,417 @@
...
+            Initializes a sectionObject which is used internally to group
+            related segments.
+            """
+            self.sectionList = []
+            self.segmentList = []
+            self.name = 'soma'
+
+    def __init__(self, file_obj, name_heuristic=True):
+        """
+        Initializes NmlMorphology Class
```
mstimberg: Here and in a few other places: "NmlMorphology" --> "NMLMorphology"

brian2tools/tests/testScript.py
```diff
@@ -0,0 +1,62 @@
+import argparse
+import os, sys
+from brian2tools.nmlimport.nml import NmlMorphology
```
mstimberg: This is now broken I guess (`NmlMorphology` --> `NMLMorphology`)

docs_sphinx/user/nmlimport.rst
```diff
@@ -0,0 +1,157 @@
...
+
+As a demonstration, we will use the nmlimport library to generate morphology and
+extract other related information from the ``pyr_4_sym.cell.nml`` nml file. You
+can download it from `here <https://github
+.com/OpenSourceBrain/ACnet2/blob/master/neuroConstruct/generatedNeuroML2/pyr_4_sym.cell.nml>`_.
+
+
+.. code:: python
+
+    from brian2tools.nmlimport.nml import NmlMorphology
```
mstimberg: `NmlMorphology` --> `NMLMorphology` (here and below in a few places)

kapilkd13:
Hi @mstimberg I was developing function to convert str values like `cond_density` to quantity object and then I found https://github.com/brian-team/brian2tools/blob/master/brian2tools/nmlexport/supporting.py#L12 this method :smiley: . This converts our string value to quantity object. I tested with few values, result looks good. We can make a copy of that function and extend it for our usecase. What do you think?
Something like this does the work:
```
from brian2tools.nmlexport.supporting import from_string
m=NMLMorphology('.nml file')
q=from_string(m.doc.cells[0].biophysical_properties.membrane_properties.channel_densities[0].cond_density)
```

mstimberg:
> Hi @mstimberg I was developing function to convert str values like cond_density to quantity object and then I found https://github.com/brian-team/brian2tools/blob/master/brian2tools/nmlexport/supporting.py#L12 this method smiley . This converts our string value to quantity object. I tested with few values, result looks good. We can make a copy of that function and extend it for our usecase. What do you think?

Ha, I completely forgot about this function, I thought we only had the conversion in the other direction! We need to test this function more thoroughly, but in principle it should do what we want. So indeed no need to write this function again. Maybe the best approach would be to have a new package in parallel to `nmlexport` and `nmlimport` that includes functions that are useful for both, something like `nmlutils`.

PR #18, opened 2018-06-06
Sphinx fix: fixes the sphinx doc build issue

kapilkd13:
Currently sphinx uses sphinx-issuetracker extension, which is out of date now. Changing it to extlinks


thesamovar:
Oops, probably @mstimberg should check this too before merging!

mstimberg:
Hah, no worries. @kapilkd13 thanks for the changes, looks all good to me. I made a small change because our call to `sphinx-apidoc` was no longer compatible with sphinx versions >= 1.7 -- let's see whether it works fine on the test servers as well.

kapilkd13:
@mstimberg  As per logs `pytz` is not being installed. I looked for its feedstock on conda-forge and it is present. what can be the reason here?

mstimberg:
> @mstimberg As per logs pytz is not being installed. I looked for its feedstock on conda-forge and it is present. what can be the reason here?

I think there's some issue with the sphinx package in the default channel, i.e. not in conda-forge. I changed `travis.yml` to prefer the conda-forge channel, let's see whether this fixes it.

PR #19, opened 2018-07-16
[WIP Enhancing nmlimport library

kapilkd13:
@mstimberg  As suggested I am moving shared functions to nmlutils library. Also further changes will be pushed on this PR.


kapilkd13:
@mstimberg  I was thinking about morphlogy properties and how should they be available to nmlimport's users. libneuroML already creates a `membraneProperties` (and other similar) objects with value as string. I was thinking to replace this string value with Brian's `quantity` object and thus only updating the necessary fields in lineuroml object and use that. What do you think?

mstimberg:
Just reusing everything from the neuroml object except for replacements of the quantities is a good basic solution, and can certainly be useful. A few things in the membrane properties and intracellular properties directly match to `SpatialNeuron` options, though, it would be good to have an easy way of re-using them. What I'm thinking of: In the `sample1.cell.nml` example, we have:
```xml
<spikeThresh value="0 mV"/>
<specificCapacitance value="2.84 uF_per_cm2"/>
...
<resistivity value="0.2 kohm_cm"/>
```
These could be set in Brian as:
```Python
neuron = SpatialNeuron(..., threshold='v > 0*mV', refractory='v > 0*mV',
                       Cm=2.48*uF/cm**2, Ri=0.2*kohm*cm)
```
(The `refractory` keyword is to implement this statement from the NeuroML spec: "Note, usually the spiking event will not be emitted again until the membrane potential has fallen below this value and rises again to cross it in a positive direction "). If `threshold`, `refractory`, `Cm`, `Ri` are exposed as a dictionary `nml_object.properties`, then a concise way to use them in Brian would be:
```Python
neuron = SpatialNeuron(..., **nml_object.properties)
```
This would be kind of nice I think :) Can you have a look, does the NeuroML spec require the same value of Ri and Cm for all compartments, or can this value differ between segments? And if it can, does this actually happen in the example files?

I'd also like to go into more details for all the information about channels, but let's discuss this another time. Maybe you could already look into what is needed to properly support `<include href='...'>` tags, though?

kapilkd13:
@mstimberg  I checked NeuroML specs and as per that  Ri and Cm values are applied on `group of sections`. So I guess it is constant for a SegmentGroup. 
I was working on properties dictionary and as per docs Ri, Cm and Threshold are provided to us as a list of objects in our `nml_object`, where each object contains `SegmentGroup` id and the parameter values in string. So I am thinking of creating a dictionary for each of the `Cm,Ri, Thresh` of the form 
`{ segment_group:value,...  }` This will allow us to later fetch these values corresponding to any group easily. What do you think?

kapilkd13:
@mstimberg  I am little confused between nml's `spikeThresh` and Brian's threshold. Is spikeThresh equivalent to threshold or refractory. And are the threshold and refractory related?

mstimberg:
Hi. Just quickly about your last question: `spikeThresh` is equivalent to the `threshold`. The refractoriness period defines how long after a spike a neuron should not be allowed to spike again. See http://brian2.readthedocs.io/en/stable/user/refractoriness.html#defining-the-refractory-period and the remark about Hodgkin-Huxley type models for why it makes sense to use the same value for the refractoriness as for the threshold in biologically detailed models.

mstimberg:
About `Cm` and `Ri`: In Brian, the value for `Ri` has to be the same for all compartments, but `Cm` can vary. However, a varying `Cm` cannot be specified as part of the `__init__` call.

My proposal with `properties` attribute was maybe a bit too simplistic, but I think it's important to make the default use case -- constant `Ri`/`Cm` over the neuron -- particularly easy. My new suggestion:
* The `NMLMorphology` object should have attributes using the Brian names, i.e. `Ri`, `Cm` and `threshold`
* For convenience, the `NMLMorphology` object should also have an attribute `threshold_string`, which is e.g. `'v > -20*mV'` if `threshold` is `20*mV`
* If the morphology only specifies a single value for the whole morphology, then this attribute should have this value (i.e. be a `Quantity`) -- this will always be the case for `threshold`.
* It the morphology defines values per segment group, then the attribute should be a dictionary as you suggested

This would mean that in the simplest case (constant values over morphology), the definition in Brian would still look reasonably simple:
```Python
neuron = SpatialNeuron(..., Cm=nml_object.Cm, Ri=nml_object.Ri, 
                       threshold=nml_object.threshold_string,
                       refractory=nml_object.threshold_string)
```

For values that vary over the morphology, it would be good to have a convenience function to apply values from a dictionary. What I mean is something like the following (assuming that `Cm` has different values for different segment groups now):
```Python
set_for_segment_groups(neuron.Cm, nml_object.Cm)
```
This would save the user a bit of typing, i.e. they won't have to type this:
```Python
for segment_group, value in nml_object.Cm.items():
    neuron.Cm[nml_object.resolved_grp_ids[segment_group]] = value
```

kapilkd13:
Hi @mstimberg What should I pass as `model` param in `SpatialNeuron` when creating this neuron from properties given.

mstimberg:
> Hi @mstimberg What should I pass as model param in SpatialNeuron when creating this neuron from properties given. For now I am returning a partial function.

Hi @kapilkd13 You do not have to create the `SpatialNeuron` yourself, this would still be up to the user. What I meant was to give the user everything they need to quickly do it themselves. In my previous comment, the `neuron = SpatialNeuron(...)` lines were supposed to be part of the user's script, not of `brian2tools`.

mstimberg:
About `ionChannel` tags: It would be great if we could support certain kind of "simple" channels explicitly by returning Brian `Equations` objects for them. What I have in mind: `NMLMorphology` would have a method `get_channel_equations` which takes the name of the `ionChannel`as an argument and returns the respective `Equation` object (or raises `NotImplementedError` if it is of an unsupported type).

The two basic types that we should support are `ionChannelPassive` and `ionChannelHH` (sometimes called  just `ionChannel` -- apparently both have identical functionality and one will be removed in the future) where `forwardRate` and `reverseRate` use one of the following types: `HHExpRate`, `HHSigmoidRate`, or `HHExpLinearRate`.

For `ionChannelPassive`, the equations that should be returned are:

`Equations('I = g/area*(erev - v) : amp/meter**2')`

`g` would be replaced by the `conductance` value from the `ionChannel` description, and `erev` would be set from the `channelDensity` tag. For now, I would simply assume/require that only the `condDensity` and not the `erev` value changes if more than one `channelDensity` exists for a given `ionChannel` (this should almost always be the case). To avoid clashes with other current definitions, the name `I` should be replaced by a name that is suffixed with the ion channel name. E.g. in our example file, `I` should be `I_LeakConductance_pyr`. Both the replacement of names and the insertion of values can be directly done with the `Equations` object. Using the information from our example file, `get_channel_equations('LeakConductance_pyr')` should return `Equations('I = g/area*(erev - v) : amp/meter**2', I='I_LeakConductance_pyr', g=10*psiemens, erev=-66*mV)`. This will result in `Equations('I_LeakConductance_pyr = 10.0*psiemens/area*(-66.0*mvolt - v) : amp/meter**2')` which is what we want.

For `ionChannelHH`/`ionChannel` , things are more complicated.

The equation consists of several parts (multi-line string). The first part is the current, similar to the passive channel (but note that `g` is no longer divided by the area -- the passive channel has a fixed total conductance, whereas the other channels give conductance *densities*) :

`I = g*...*(erev - v) : amp/meter**2`

The `...` part is new: here, we'll have to multiply all gating variables, as many times as their `instances` value says. E.g. for the [`Na_pyr`](https://github.com/OpenSourceBrain/ACnet2/blob/master/neuroConstruct/generatedNeuroML2/Na_pyr.channel.nml) channel, the channel defines a gating variable `m` with `instances=2` and `n` with `instances=1`. This leads to:

`I = g*m*m*n*(erev - v) : amp/meter**2`

Or written differently:

`I = g*m**2*n*(erev - v) : amp/meter**2`

The next part is one differential equation for each of the gating variables, using appropriate names/suffixes (but apart from that identical), e.g. for the gating variable `m`:

`dm/dt = alpha_m*(1-m) - beta_m*m : 1`

`alpha` stands for what NeuroML calls the `forwardRate` and `beta` for the `reverseRate`.

Then, we need equations for those:

```
alpha = ... : second**-1
beta = ... : second**-1
```

The `...` depends on the type of the `forwardRate` and `reverseRate`. The equations are what is given in the [NeuroML docs](https://www.neuroml.org/NeuroML2CoreTypes/Channels.html) for `r`. This means, use `rate * exp((v - midpoint)/scale)` for `HHExpRate` , and `rate / (1 + exp(0 - (v - midpoint)/scale))` for `HHSigmoidRate` . 

The equation for `HHExpLinearRate` should be more complicated to avoid a division by 0, but as a first step, we can ignore the fact that it is not well-defined if  `v` happens to have exactly the same value as `midpoint` . So let's just use `rate * (v - midpoint) / scale / (1 - exp(- (v - midpoint) / scale))`

Finally, in contrast to the passive channel, the conductance is not constant but normally varies across the neuron. We therefore have to add it as a neuron parameter to the equations:

```
g : siemens/meter**2
```

As you see, this gets quite long and complicated fast, but it is only a question of using the right template and plugging all the names/values in. Here's what `get_channel_equations('Na_pyr')` should return in the end (i.e. with names/values already inserted). I did this by hand, so it is very possible that there is some mistake somewhere, though!

```python
Equations('''
I_Na_pyr = g_Na_pyr*m_Na_pyr*m_Na_pyr*n_Na_pyr*(55.0*mV - v) : amp/meter**2

dm_Na_pyr/dt = alpha_m_Na_pyr*(1-m_Na_pyr) - beta_m_Na_pyr*m_Na_pyr : 1
# forwardRate for m is HHExpLinearRate
alpha_m_Na_pyr = 1.28e3/second * (v - -4.69e-2*volt) / 0.004*volt / (1 - exp(- (v - -4.69e-2*volt) / 0.004*volt)) : second**-1
# reverseRate for m is HHExpLinearRate
beta_m_Na_pyr = 1.4e3/second * (v - -19.9e-3*volt) / -0.005*volt / (1 - exp(- (v - -19.9e-3*volt) / -0.005*volt)) : second**-1

dn_Na_pyr/dt = alpha_n_Na_pyr*(1-n_Na_pyr) - beta_n_Na_pyr*n_Na_pyr : 1
# forwardRate for n is HHExpRate
alpha_n_Na_pyr = 1.28e2/second * exp((v - -4.3e-2*volt)/-0.018*volt) : second**-1
# reverseRate for n is HHSigmoidRate
beta_n_Na_pyr = 4e3/second / (1 + exp(- (v - -2e-2*volt)/0.005*volt)) : second**-1

g_Na_pyr : siemens/meter**2
''')
```

Hope that makes somewhat sense and gets you going!

kapilkd13:
Hi @mstimberg Everthing is going fine till now. Currently I am working on `ionChannel` part. The values present in channel files have different format than the values in morphology file. So I am modifying `string_to_quantity` method to address this. I will ping you once I am done with it.
I have one little doubt though,

> Equations('''
I_Na_pyr = g_Na_pyr*m_Na_pyr*m_Na_pyr*n_Na_pyr*(55.0*mV - v) : amp/meter**2
dm_Na_pyr/dt = alpha_m_Na_pyr*(1-m_Na_pyr) - beta_m_Na_pyr*m_Na_pyr : 1
alpha_m_Na_pyr = 1.28e3/second * (v - -4.69e-2*volt) / 0.004*volt / (1 - exp(- (v - -4.69e-2*volt) / 0.004*volt)) : second**-1

In this string there are values like `1.28e3/second ` These have slightly different notation in nml file. Can I directly put string values present in nml file in this string. Currently I am creating a quantity object from these string values, In `Equation` I am writing string variables and later passing these quantity object as named param to Equation function. for ex here https://github.com/brian-team/brian2tools/pull/19/commits/01e93dea79534cab913572b6357ae88a096e095d#diff-e3ef6f9d86b4cf8442c5aa096645fbffR547
So which way would be preferred?

kapilkd13:
Hi @mstimberg  Here is a link to [my final report](https://gist.github.com/kapilkd13/dd6ea54af8091441832e536cd48bb758). Please have a look and suggest any changes if you want.

mstimberg:
Hi @kapilkd13 , hope you saw my comments on the gist? I had a look at the code as well and everything looks good -- if you are still motivated, there are a few stylistic PEP8 violations that you might want to correct, though ;)

I'll do a more thorough test with an actual example before merging this PR, but I don't see anything significant to do from your side anymore.

kapilkd13:
Hi @mstimberg  I have made the necessary changes suggested by you in the final report gist. You may like to have a final look. 
I will make PEP8 corrections later today.

mstimberg:
> I will make PEP8 corrections later today.

Don't worry about those corrections, I'll make a number of small changes anyway and can correct them in passing.

kapilkd13:
Cool. :smiley_cat:   I did few pep 8 corrections though. Its not much, just some import library ordering and few line spaces. Should I push it or you would do this.

mstimberg:
I'm not working on it right now, so please feel free to push it :)

kapilkd13:
Hi @mstimberg I am free this weekend, so if there are any changes you want I can do it. Would love to get it merged soon. :smile:

mstimberg:
Hi @kapilkd13 , sorry, I completely overlooked your message! The reason I've not merged this is that I noticed that some NeuroML files are formulated a bit differently than what the current code supports. 

If you have a look at https://github.com/NeuroML/NeuroML2/blob/master/examples/NML2_SingleCompHHCell.nml for example, it is a type of cell that we should support. The syntax is a bit different, though. E.g. instead of
```xml
<ionChannel id="Na_pyr" conductance="10pS" type="ionChannelHH" species="na">
```
(which is what your code expects and supports) it uses
```xml
<ionChannelHH id="naChan" conductance="10pS" species="na">
```
which should mean the same thing. I wanted to read up on this first, but I did not find the time yet...

mstimberg (approved):
This took an awful long time, but I'll finally go ahead and merge these latest changes. I will also prepare a release soon, to finally make all this easily available to all users.
Thanks again for all your work on this @kapilkd13 !

PR #31, opened 2020-06-16
`ExportDevice` mode to export standard representation

### LINKED ISSUE #26: Create a new general export device
For GSoC 2020: a new device (similar to `LemsDevice`) to generate a general representation of a Brian model.

Vigneswaran-Chandrasekaran:
Class `StdDevice` that derives `brian2.devices.device.RuntimeDevice` to run the Brian script in `stdformat` device mode to generate the standard dictionary representation. 

To set `stdformat` device mode,
```python
from brian2tools import stdformatexport

set_device('stdformat')
```
It shall be used to access/collect namespace, variables and other information that are non-group specific. 
Fixes #26


Vigneswaran-Chandrasekaran:
Simple example for using `stdformat` device,
```python
from brian2tools import stdformatexport

set_device('stdformat')

eqn = 'dv/dt = (1 - v)/tau :1'
tau = 10 * ms

grp = NeuronGroup(1, eqn, method = 'exact')

run(100 * ms)
```
The output would be (stored in self.network_dict),
```
{'neurongroup': 
[{'name': 'neurongroup',
                            'N': 1,
                            'user_method': 'exact', 
                            'user_equations': {'v': {'expr': Expression('(1 - v)/tau'), 
                                                         'unit': radian, 
                                                         'type': 'differential equation', 
                                                         'dtype': 'float'}
                                                         }
}]
}
```

mstimberg:
To get the external identifiers resolved in the same way as Brian does it, provide the list of names together with the run namespace to the [`Group.resolve_all`](https://brian2.readthedocs.io/en/stable/reference/brian2.groups.group.Group.html#brian2.groups.group.Group.resolve_all) function. Classes like `NeuronGroup` all inherit from `Group`, so they have access to this method.
You can dump *all* the names (names of functions/variables/constants/...) in it, but then filter the output a bit since we don't need e.g. `ArrayVariable` objects. I think what should be represented are `Constant` objects (which corresponds to any external variable used in the code) and probably also `Function` objects (to deal with user-defined functions or e.g. `TimedArray`). In both cases, you can discard them if these are functions or constants that are also part of `DEFAULT_FUNCTIONS` or `DEFAULT_CONSTANTS`, respectively. If equations refer to e.g. `exp` there's no need to explicitly store what this represents. Same for a reference to `pi`, for example.

Vigneswaran-Chandrasekaran:
Simple example use_case of `ExportDevice` mode.
```python
from brian2 import *
from brian2tools import baseexport

set_device('ExportDevice', build_on_run = False)
eqn = '''dv/dt = (v0 - v) / tau :volt'''
tau = 10 * ms
v0 = 1 * volt
grp = NeuronGroup(10, model = eqn, method = 'exact', 
                  threshold = 'v > 100 * mV', 
                  reset = 'v = 0 * mV', events = {'custom_event': 'v > 2*volt'})
grp.run_regularly('v = -10 * mV', dt = 10 * ms)
po  = PoissonGroup(10, rates = 100 * Hz)
mon = StateMonitor(grp, 'v', record = True)
spmon = SpikeMonitor(po, record = True)
evmon = EventMonitor(grp, 'custom_event', record = True)
run(100 * ms)

chk_var = 2 * mV
chk_var2 = -10 * mV
grp.v[2:5] = 'chk_var2'
run(200 * ms)
device.build(debug = True)
```
Output,
```
WARNING:brian2tools.baseexport:The package is under development and may give incorrect results
[{'components': {'eventmonitor': [{'dt': 100. * usecond,
                                   'event': 'custom_event',
                                   'name': 'eventmonitor',
                                   'order': 1,
                                   'record': True,
                                   'source': 'neurongroup',
                                   'variables': ['i', 't'],
                                   'when': 'after_thresholds'}],
                 'neurongroup': [{'N': 10,
                                  'equations': {'v': {'expr': '(v0 - v) / tau',
                                                      'type': 'differential '
                                                              'equation',
                                                      'unit': volt,
                                                      'var_type': 'float'}},
                                  'events': {'spike': {'reset': {'code': 'v = '0 * mV',
                                                                 'dt': 100. * usecond,
                                                                 'order': 0,
                                                                 'when': 'resets'},
                                                       'threshold': {'code': 'v > 100 * mV',
                                                                     'dt': 100. * usecond,
                                                                     'order': 0,
                                                                     'when': 'thresholds'}},
                                             'custom_event': {'threshold': {'code': 'v > 2*volt',
                                                                           'dt': 100. * usecond,
                                                                           'order': 0,
                                                                           'when': 'after_thresholds'}}},
                                  'identifiers': {'mV': 1. * mvolt,
                                                  'tau': 10. * msecond,
                                                  'v0': 1. * volt,
                                                  'volt': 1. * volt},
                                  'name': 'neurongroup',
                                  'order': 0,
                                  'run_regularly': [{'code': 'v = -10 * mV',
                                                     'dt': 10. * msecond,
                                                     'name': 'neurongroup_run_regularly',
                                                     'order': 0,
                                                     'when': 'start'}],
                                  'user_method': 'exact',
                                  'when': 'groups'}],
                 'poissongroup': [{'N': 10,
                                   'name': 'poissongroup',
                                   'rates': 100. * hertz}],
                 'spikemonitor': [{'dt': 100. * usecond,
                                   'event': 'spike',
                                   'name': 'spikemonitor',
                                   'order': 1,
                                   'record': True,
                                   'source': 'poissongroup',
                                   'variables': ['i', 't'],
                                   'when': 'thresholds'}],
                 'statemonitor': [{'dt': 100. * usecond,
                                   'n_indices': 10,
                                   'name': 'statemonitor',
                                   'order': 0,
                                   'record': True,
                                   'source': 'neurongroup',
                                   'variables': ['v'],
                                   'when': 'start'}]},
  'duration': 100. * msecond,
  'initializers': [{'index': True,
                    'source': 'poissongroup',
                    'value': 100. * hertz,
                    'variable': 'rates'}]},
{ ..
'initializers': [{'index': array([2, 3, 4], dtype=int32),
                    'source': 'neurongroup',
                    'value': 'chk_var2',
                    'variable': 'v'}],
run_identifiers': {'chk_var2': -10. * mvolt}}]
```

mstimberg:
I pushed two commits directly to your branch, hopefully they should fix the documentation errors. The main issue was that we are "mocking" the Brian2 modules, so that you can build the documentation for brian2tools without having Brian2 installed (not super-important in practice, but it speeds up the whole process on readthedocs). Unfortunately, this means manually putting the names of all modules that are used somewhere in the code into `conf.py` – in this case, `brian2.core.functions` wasn't present and led to the error.

Vigneswaran-Chandrasekaran:
Great! Thank you very much :smile: 
Regarding the representation of identifiers from initializers, I made the changes to make the dictionary to look something like,
```
'initializers': [{'index': array([2, 3, 4], dtype=int32),
                    'source': 'neurongroup',
                    'value': 'chk_var2',
                    'variable': 'v',
                    'identifiers': {'chk_var2': -10. * mvolt}
                    }]
```

PR #34, opened 2020-06-25
Add collectors for Monitors and run_regularly()

### LINKED ISSUE #32: Represent monitors in the description dictionary
Also represent `StateMonitor`, `SpikeMonitor` and `PopulationRateMonitor` in the structure created as part of #30

### LINKED ISSUE #33: Represent run_regularly operations in the description dictionary
Also represent `run_regularly` operations in the structure created as part of #30

Vigneswaran-Chandrasekaran:
Fixes #32 
Fixes #33 

Add functions namely, `collect_StateMonitor()`, `collect_SpikeMonitor()` and `collect_PopulationRateMonitor`to get information from `StateMonitor`, `SpikeMonitor` and `PopulationRateMonitor` object respectively.

### DIFF
--- a/brian2tools/baseexport/collector.py
+++ b/brian2tools/baseexport/collector.py
@@ -122,20 +122,26 @@ def collect_Events(group):
         Dictionary with extracted information
     """
 
-    event_dict = {}
+    spike_dict = {}
 
     # add threshold
-    event_dict['spike'] = {'threshold': group.events['spike']}
+    spike_dict['threshold'] = {'code': group.events['spike'],
+                               'when': group.thresholder['spike'].when,
+                               'order': group.thresholder['spike'].order,
+                               'dt': group.thresholder['spike'].clock.dt}
 
     # check reset is defined
     if group.event_codes:
-        event_dict['spike'].update({'reset': group.event_codes['spike']})
+        spike_dict.update({'reset': {'code': group.event_codes['spike'],
+                                     'when': group.resetter['spike'].when,
+                                     'order': group.resetter['spike'].order,
+                                     'dt': group.resetter['spike'].clock.dt}})
 
     # check refractory is defined
     if group._refractory:
-        event_dict['spike'].update({'refractory': group._refractory})
+        spike_dict.update({'refractory': group._refractory})
 
-    return event_dict
+    return {'spike': spike_dict}
 
 
 def collect_SpikeGenerator(spike_gen):
--- a/brian2tools/baseexport/collector.py
+++ b/brian2tools/baseexport/collector.py
@@ -47,7 +47,6 @@ def collect_NeuronGroup(group):
         neuron_dict['events'] = collect_Events(group)
 
     # check any `run_regularly` / CodeRunner objects associated
-
     for obj in group.contained_objects:
         # Note: Thresholder, StateUpdater, Resetter are all derived from
         # CodeRunner, so to identify `run_regularly` object we use type()
@@ -193,6 +192,17 @@ def collect_PoissonGroup(poisson_grp):
 
     # get rates (can be Quantity or str)
     poisson_grp_dict['rates'] = poisson_grp._rates
+
+    # `run_regularly` / CodeRunner objects of poisson_grp
+    for obj in poisson_grp.contained_objects:
+        if type(obj) == CodeRunner:
+            if 'run_regularly' not in poisson_grp_dict:
+                poisson_grp_dict['run_regularly'] = []
+            poisson_grp_dict['run_regularly'].append({
+                                                'name': obj.name,
+                                                'code': obj.abstract_code,
+                                                'dt': obj.clock.dt
+                                                })
 
     return poisson_grp_dict
 
--- a/brian2tools/baseexport/collector.py
+++ b/brian2tools/baseexport/collector.py
@@ -122,26 +122,30 @@ def collect_Events(group):
         Dictionary with extracted information
     """
 
-    spike_dict = {}
+    event_dict = {}
+    # loop over the thresholder to check `spike` or custom event
+    for event in group.thresholder:
+        # for simplicity create subdict variable for particular event
+        event_dict[event] = {}
+        event_subdict = event_dict[event]
+        # add threshold
+        event_subdict['threshold'] = {'code': group.events[event],
+                                      'when': group.thresholder[event].when,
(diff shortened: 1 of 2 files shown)


Vigneswaran-Chandrasekaran:
Collecting functions for Monitors and `run_regularly()` for `NeuronGroup` have been added along with test cases. 

I believe the PR is now ready for review.

Vigneswaran-Chandrasekaran:
Yes, that would be good and I remember you told me about that before. 

To get `when` and `order` attributes from `Group` based objects, we would directly collect its own and iterate over their `CodeRunner` objects (in `.contained_objects`). Is that a good way?

mstimberg:
It can sometimes be a bit tricky, because objects such as `NeuronGroup` have `when` and `order`, but it is not used – the `NeuronGroup` itself is not doing anything during a run, all work is delegated to the `StateUpdater`, `Thresholder` and `Resetter` objects. In this case, we do not have to store the `when`/`order` attribute of the `NeuronGroup`, but only of the other (contained) objects. In the dictionary, I would probably treat the group as being equivalent to its `StateUpdater`, i.e. add the state updater's `when`/`order` attributes at the top level of the `NeuronGroup`.

You can recognize this situation by the fact that `NeuronGroup` does not accept a `when` argument in its initializer – it wouldn't make sense since the `NeuronGroup` creates `StateUpdater`, `Thresholder`, and `Resetter`, which all use their own `when` slot. On the other hand, it allows the specification of `order`, but this is just passed down to the other objects. 

I think of the objects you are supporting so far `NeuronGroup` is the only difficult case (`Synapses` would be the other one), for `PoissonGroup` etc. you'll always have a single `when`/`order` attribute (and an additional pair for `run_regularly`, if it exists).

Vigneswaran-Chandrasekaran:
Thanks for the clear explanation :) 
I'll make the changes accordingly

Vigneswaran-Chandrasekaran:
Added `when` and `order` attributes in the dictionary representation. `NeuronGroup`'s `when`/`order` is set with `StateUpdater`'s. However, I've not added `when` and `order` for other `.contained_objects` like `Thresholder`, `Resetter` etc. (except `run_regularly`), as their `when` slots are going to be fixed (like for `Thresholder`: `thresholds`, `Resetter`: `resets`) and `order` is same as source's `order`. If these changes are good to go, then it is ready for your review.

mstimberg:
For `Thresholder` and `Resetter`, your reasoning makes perfect sense. However, it *is* possible to change their `when` and `order` attributes, if you directly set it on `group.thresholder`, etc. after creating the `NeuronGroup` object. I don't think we mention this anywhere in the documentation, but I am sure I have mentioned it on the mailing list for specific issues, so I think it would be better to support it. We also have a (very old...) open issue (brian-team/brian2#790) about making it easier to set these attributes directly, so it might become more relevant in the future.

Vigneswaran-Chandrasekaran:
Oh, then representing something like this would be good option?
```
{
...
'spike`:
   {'threshold': 
     {'code' : . . . ,
      'when' : . . .,
      'order' : . . .,
      'dt' : . . . 
     }
     . . . . 
   }
. . . .
}
```

mstimberg:
Yes!

Vigneswaran-Chandrasekaran:
I've made the changes for adding `when`, `order` and `dt` for `Thresholder` and `Resetter` objects and respective test cases.

mstimberg:
Hmm, if you prefer, we can also deal with this in a new PR. I just realized that it goes a bit beyond what the initial aim of this PR was.

Vigneswaran-Chandrasekaran:
Hi, thanks for the review :)

I too realized that hardcoding `spike` with the event dictionary is not looking good, but not sure of how to handle custom events with the same `collect_Events()`. If supporting custom events with current code shall be done with few changes in the current code (like refractory check for `spike` and removing 'spike' attribute as you mentioned), then I guess this PR wouldn't look much deviated from initial aim. However, handling custom events is an important thing and indeed would be better to deal with a separate PR.

mstimberg:
It should be rather straightforward as written in my review: you basically use your code but put a loop with `for event in group.thresholder` around it, and replace your mentions of `'spike'` by `event`. If (like in most cases), the group only has the standard `spike` event, the output will be exactly the same as now (tests don't have to change). But if there are other events, they would be represented as well under their own name. The only other thing to change would be to replace:
```Python
if group._refractory:
```
by
```Python
if event == 'spike' and group._refractory:
```

But if you prefer to deal with this later, please let me know and I merge the branch as it is.

Vigneswaran-Chandrasekaran:
Thank you for the clear answer :relaxed: I think, we shall work with the same PR as the changes are looking straightforward
A couple of small doubts are,
- Can we add collector functions for `EventMonitor` too? such that monitor part for custom events shall be taken care too
- Also, shall we add the `run_on_event` field too with the event dictionary?

mstimberg:
Sure, `EventMonitor` would be good to support, too. It's all identical to `SpikeMonitor`, except that we also need to store the name of the event. For full generality, we should probably only store `EventMonitor`s in the dictionary, with `SpikeMonitor` being just one special case which monitors the `spike` event. But we could also keep them separate since most of the time, there are no `EventMonitor`s.

For the `run_on_event`: this should already be taken care of automatically! They end up in the `resetter`/`event_codes` in the same way as the code for `reset` (which is a `run_on_event` for the `spike` event).

Vigneswaran-Chandrasekaran:
Oh great then, I'll make the necessary changes!

Vigneswaran-Chandrasekaran:
I've made the changes to support custom events, collector function for `EventMonitor`, and test cases. 
With the addition of custom events to the dictionary format, the event handling collector function is better looking now. 
I believe, the PR is now ready for your kind review.

Vigneswaran-Chandrasekaran:
Yes, that is correct and it looks quite redundant. I think we shall have the second option of, calling `collect_EventMonitor()` inside `collect_SpikeMonitor()`?, If that's fine, I'll add the changes in other PR?

>I am going ahead with the merge to not block your other work

Not like that, I parallelly work in a different branch irrespective of this

Thank you very much for your review :)

mstimberg:
>  I think we shall have the second option of, calling `collect_EventMonitor()` inside `collect_SpikeMonitor()`?, If that's fine, I'll add the changes in other PR?

Sure, or feel free to include the change as part of any other PR.

PR #39, opened 2020-08-05
Supporting Synapses for Standard representation

### LINKED ISSUE #38: Represent Synapses in the description dictionary
Representing `Synapses` in standard dictionary format similar to other supported types

Vigneswaran-Chandrasekaran:
Support `Synapses` in the standard dictionary representation

Fixes #38


Vigneswaran-Chandrasekaran:
With the initial commit, the output for sample example,

```python
set_device('ExportDevice', build_on_run=False)


model_eqn = '''
dv/dt = (v_th - v)/tau :volt
v_th :volt
'''

tau = 10 * ms
P = NeuronGroup(10, model_eqn, method='euler', threshold='v > 900 * mV')
Q = NeuronGroup(10, model_eqn, method='euler', threshold='v > 700 * mV', reset = 'v = 100 * mV')

P.v_th = 1 * volt
Q.v_th = 1 * volt

eqn = '''
dp/dt = (1 - vp)/taup :1 (event-driven)
taup = 10 * ms :second
v_th_post = v :volt (summed)
w :volt
k :volt
'''

syn = Synapses(P, Q, eqn, on_pre='v += w', on_post = 'v -= w')
syn.connect(i=[0, 1, 2], j = [0, 3, 4], n=2)

run(100 * ms)
device.build(debug=True)
```
looks like (synapses part alone is shown) in the component level,
```
                'synapses': [{'equations': {'k': {'type': 'parameter',
                                                   'unit': volt,
                                                   'var_type': 'float'},
                                             'p': {'expr': '(1 - vp)/taup',
                                                   'flags': ['event-driven'],
                                                   'type': 'differential equation',
                                                   'unit': radian,
                                                   'var_type': 'float'},
                                             'taup': {'expr': '10 * ms',
                                                      'type': 'subexpression',
                                                      'unit': second,
                                                      'var_type': 'float'},
                                             'w': {'type': 'parameter',
                                                   'unit': volt,
                                                   'var_type': 'float'}},
                               'name': 'synapses',
                               'source': 'neurongroup',
                               'summed_variables': [{'code': '\n'
                                                             '        '
                                                             '_synaptic_var = '
                                                             'v\n'
                                                             '        ',
                                                     'dt': 100. * usecond,
                                                     'name': 'synapses_summed_variable_v_th_post',
                                                     'order': -1,
                                                     'target': 'neurongroup_1',
                                                     'when': 'groups'}],
                               'target': 'neurongroup_1'}]},
```
and in the run level,
```
'synaptic_connections': [{'j': 'k for k in sample(N_post, p=i*1.0/(N_pre-1))',
                            'n_connections': 2,
                            'probability': 1.0}],
  'synaptic_pathways': [{'clock': 100. * usecond,
                         'code': 'v += w',
                         'event': 'spike',
                         'name': 'synapses_pre',
                         'order': -1,
                         'prepost': 'pre',
                         'source': 'neurongroup',
                         'synapse': 'synapses',
                         'target': 'neurongroup_1',
                         'when': 'synapses'},
                        {'clock': 100. * usecond,
                         'code': 'v -= w',
                         'event': 'spike',
                         'name': 'synapses_post',
                         'order': 1,
                         'prepost': 'post',
                         'source': 'neurongroup_1',
                         'synapse': 'synapses',
                         'target': 'neurongroup',
                         'when': 'synapses'}]}]
```

Vigneswaran-Chandrasekaran:
Now, `PoissonInput`, `TimedArray`, and custom functions are also supported for standard representation.
A couple of things, to highlight are,
- In `PoissonInput`, information for `weight` and `target_var` are not directly available, therefore used regular expression to fetch the values from `CodeRunner.abstract_code`
- With `TimedArray` representation, I added them in the `identifiers` field, as it can be used in many ways like,
```python
neurongrp.v[:] = 'rand() * timedarr(2*ms)
poissongroup.rates = 'timedrates(t)`
```
and in equations (similarly custom functions also). Kindly let me know if there is any more efficient way to handle this

mstimberg:
>     * In `PoissonInput`, information for `weight` and `target_var` are not directly available, therefore used regular expression to fetch the values from `CodeRunner.abstract_code`

I now added this information in Brian master for easier extraction.



>     * With `TimedArray` representation, I added them in the `identifiers` field

Yes, that's the right place to put it I think :+1:

Vigneswaran-Chandrasekaran:
Thank you :blush: I'll make the changes accordingly

Vigneswaran-Chandrasekaran:
@mstimberg  I think, the PR is now ready for your review :)

brian2tools/baseexport/collector.py
```diff
@@ -226,6 +260,17 @@ def collect_PoissonGroup(poisson_grp):
 
     # get rates (can be Quantity or str)
     poisson_grp_dict['rates'] = poisson_grp._rates
+    if type(poisson_grp._rates) == str:
```
mstimberg, line 263: Idiomatically this should rather be `if isinstance(poisson_grp._rates, str)`, even though it won't make any difference in practice.

brian2tools/baseexport/collector.py
```diff
@@ -133,22 +158,24 @@ def collect_Events(group):
                                       'when': group.thresholder[event].when,
                                       'order': group.thresholder[event].order,
                                       'dt': group.thresholder[event].clock.dt}
+        event_identifiers |= (get_identifiers(group.events[event]))
```
mstimberg, line 161: Tiny nitpicking comment: you don't need the outer parentheses on the right-hand-side.

brian2tools/baseexport/collector.py
```diff
@@ -133,22 +158,24 @@ def collect_Events(group):
...
 
         # check reset is defined
         if event in group.event_codes:
             event_subdict['reset'] = {'code': group.event_codes[event],
                                       'when': group.resetter[event].when,
                                       'order': group.resetter[event].order,
                                       'dt': group.resetter[event].clock.dt}
+            event_identifiers |= (get_identifiers(group.event_codes[event]))
```
mstimberg, line 169: same as above

brian2tools/baseexport/collector.py
```diff
@@ -412,3 +431,127 @@ def collect_PopulationRateMonitor(poprate_mon):
...
+    pathways = []
+    for obj in synapses.contained_objects:
+        # check summed variables
+        if isinstance(obj, SummedVariableUpdater):
+            summed_var = {'code': obj.abstract_code, 'target': obj.target.name,
+                          'name': obj.name, 'dt': obj.clock.dt,
+                          'when': obj.when, 'order': obj.order
+                         }
+            summed_variables.append(summed_var)
+            # TODO: raises error because of _synaptic_var
```
mstimberg, line 488: Is this comment still relevant?

brian2tools/baseexport/collector.py
```diff
@@ -412,3 +431,127 @@ def collect_PopulationRateMonitor(poprate_mon):
...
+        synapse_dict['pathways'] = pathways
+    # resolve identifiers and add to dict
+    identifiers = synapses.resolve_all(identifiers, run_namespace)
+    identifiers = _prepare_identifiers(identifiers)
+    if identifiers:
+        synapse_dict['identifiers'] = identifiers
+    # get when and order
+    # is this necessary or CodeRunner's is sufficient?
+    synapse_dict['when'] = synapses.when
+    synapse_dict['order'] = synapses.order
```
mstimberg, line 518: The `when` is never used (I think), and the `order` is only used to set the `order` attribute of components. This is so that you can create a `Synapses(..., order=100)` for example and be sure that all its components will be executed later than another `Synapses(...)` object. But since your description contains `when`/`order` for all the components, no need to store this information here.

brian2tools/baseexport/device.py
```diff
@@ -0,0 +1,417 @@
...
+            np.set_printoptions(threshold=10)
+            if pprint_available:
+                pprint.pprint(self.runs)
+
+
+msg = "The package is under development and may give incorrect results"
+logger.warn(msg)
+# instantiate StdDevice object and add to all_devices
+std_device = BaseExporter()
+all_devices['ExportDevice'] = std_device
```
mstimberg, line 417: Our device names are generally in lower case, and we don't need "device" as part of the name. How about just `'exporter'` ?

brian2tools/baseexport/device.py
```diff
@@ -0,0 +1,417 @@
...
+                               'and "device.activate()". Note that you '
+                               'will have to set build options (e.g. the '
+                               'directory) and defaultclock.dt again.')
+        # change the flag
+        self.has_been_run = True
+        # if to operate in debug mode
+        if debug:
+            logger.debug("Building ExportDevice in debug mode")
+            # print dictionary format using pprint
+            np.set_printoptions(threshold=10)
```
mstimberg, line 408: Small thing, but it would be good to reset the print options afterwards, otherwise they are changed for the remainder of the Python session.

brian2tools/baseexport/device.py
```diff
@@ -0,0 +1,417 @@
...
+        Collect all run information and export the standard
+        dictionary format
+
+        Parameters
+        ----------
+        direct_call: bool, optional
+            To check the call to build() was made directly
+
+        debug: bool, optional
+            To build the device in debug mode
```
mstimberg, line 382: I'd add a half-sentence here saying what this actually means ("... which prints out the dictionary information")

mstimberg (changes requested):
This looks great! I made a couple of minor comments, but nothing major. I only had a cursory look at the tests, but they seem to cover things quite nicely.

brian2tools/baseexport/collector.py
Vigneswaran-Chandrasekaran: Yes, thank you :) now it is clear. (I had that in my list of "to be cleared with you" and forgot)

brian2tools/baseexport/device.py
Vigneswaran-Chandrasekaran: yes, that is good. Also, should there be any specific logger statements when using the device?, currently we have something like,
`The package is under development and may give incorrect results`

mstimberg: To be honest, I think we can simply remove this warning. This is already quite well-tested and while there are certainly bugs, I rather expect them to lead to errors than to "incorrect results". We'll do some more testing for the final release, though :)

Vigneswaran-Chandrasekaran: yes, that sounds great :) I'll do it accordingly

Vigneswaran-Chandrasekaran:
I made the changes and ready for your kind perusal

mstimberg (changes requested):
This all looks great! I ran it on some complex examples and it... just worked :smile:

I noticed a small problem that I did not realize earlier, though: the `run_regularly` operations also need to have their referenced `identifiers`. The code should be very similar to what you do elsewhere I hope!

Vigneswaran-Chandrasekaran:
Thank you very much for pointing it out :relaxed:, I made the changes and updated test cases too

PR #40, opened 2020-09-02
Add documentations

Vigneswaran-Chandrasekaran:
Documentation for base exporter and markdown exporter


mstimberg:
Did not look at the details yet, but the general structure looks good to me!
Could you remove the files in `docs_sphinx/reference`? They will be generated automatically when the documentation is built. Since they are in the ".gitignore" file, the following should get rid of them (from [here](https://stackoverflow.com/questions/13541615/how-to-remove-files-that-are-listed-in-the-gitignore-but-still-on-the-repositor)):
```bash
git rm -r --cached .
git add .
git commit -m "Removing all files in .gitignore"
```

Vigneswaran-Chandrasekaran:
Yes sure, thank you :blush: I'll make the required changes!

Vigneswaran-Chandrasekaran:
I think the PR is ready for your kind review

mstimberg:
This is all looking good to me content-wise, love the table example for the custom expander!

I merged current master into the branch, because otherwise it is difficult to check whether the links to the API reference work. This is not everywhere the case at the moment, in some places you'd have to replace things like ```` `mdexpander` ```` by ```` `.mdexpander` ```` or ```` `~brian2tools.mdexpander` ````. Also there are a few places where I think the single and the double backtick are not used correctly (this is always confusing me when using RST instead of markdown...): Single backticks should be reserved for links to the API reference, and double ticks for everything else that should be written in typewriter font. E.g. instead of
```
For instance, `identifiers` and `dt` fields have values of type `Quantity` but `N` (population size)
is of type `int`.
```
you'd write
```
For instance, ``identifiers`` and ``dt`` fields have values of type `Quantity` but ``N`` (population size)
is of type ``int``.
```
(I am not sure whether it is really worth the effort, but to get the link into the Brian2 documentation right, you'd have to give the fully qualified name of the class, i.e. ```` `~brian2.units.fundamentalunits.Quantity` ````).

If you run sphinx in "nit-picky mode" (`-n`), it will complain about things like ``` `N` ```.

There are also a few places where you mention another section or document. To correctly link those you need to use [`:ref:`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-ref) or [`:doc:`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-doc), respectively.

Vigneswaran-Chandrasekaran:
Great! Thank you :relaxed: I'll do the changes

Vigneswaran-Chandrasekaran:
Thank you for the useful suggestions. 
I made the changes and hope it is now ready for your kind review!

mstimberg (approved):
I corrected a few minor things, but overall it is looking great now. I will probably still smoothen a few language issues at some point, but I think everything is perfectly understandable as it is.

A minor thing that I corrected that might be worth mentioning: you don't actually have to do
```Python
set_device('exporter', build_on_run=False)
...
run(...)
device.build(debug=True)
```
because any additional options to the `set_device` function will be automatically handed over to the `build` function. The following therefore works and is a bit simpler:
```Python
set_device('exporter', debug=True)
...
run(...)
```

PR #42, opened 2020-10-14
Refactor NeuroML export 

Vigneswaran-Chandrasekaran:
Refactor `nmlexport` using `baseexport`


mstimberg:
Hi @Vigneswaran-Chandrasekaran. I don't quite remember: was there still much to do for this refactoring? If not it would be great to finish it soonish so it can be part of the next release. Thanks!

Vigneswaran-Chandrasekaran:
Hi, sorry for the delay. 
I've made the changes to use `baseexport` to get model information. It works and would able to replicate the previous version of `nmlexport`, but I think it would be good to get some quick and brief comments on the changes from you before the review.

mstimberg:
The general approach looks good, this is what I had in mind. I did not quite delve into the details yet, but currently the NeuroML export does not quite result in the same NeuroML file as it did before (and the new one does not seem to be valid NeuroML). For example, if you use the IF_curve_LIF.py example (like in the [documentation](https://brian2tools.readthedocs.io/en/stable/user/exporter_user.html#working-example)), this now leads to
```xml
  <ComponentType name="neurongroup" extends="baseCell">
    <Property name="tau" dimension="time"/>
    <EventPort name="spike" direction="out"/>
    <Exposure name="v" dimension="voltage"/>
    <Dynamics>
      <StateVariable name="v" dimension="voltage" exposure="v"/>
      <StateVariable name="v0" dimension="voltage"/>
      <OnStart>
        <StateAssignment variable="v" value="0"/>
        <StateAssignment variable="v0" value="((20 * mV) * i) / (n - 1)"/>
      </OnStart>
  ...
  <ComponentType name="neurongroupMulti" extends="basePopulation">
    <Parameter name="N" dimension="none"/>
    <Parameter name="v_p" dimension="voltage"/>
    <Structure>
      <MultiInstantiate componentType="neurongroup" number="N">
        <Assign property="v" value="v_p"/>
      </MultiInstantiate>
    </Structure>
  </ComponentType>
```
which will not work with the references to `i` and `n` in the `StateAssignment` for `v0`. I only half understand this `MultiInstantiate` thing myself, but previously it generated this (which worked):
```xml
<ComponentType extends="baseCell" name="neuron1">
    <Property dimension="voltage" name="v0"/>
    <Property dimension="time" name="tau"/>
    <EventPort direction="out" name="spike"/>
    <Exposure dimension="voltage" name="v"/>
    <Dynamics>
      <StateVariable dimension="voltage" exposure="v" name="v"/>
      <OnStart>
        <StateAssignment value="0" variable="v"/>
      </OnStart>
...
  <ComponentType extends="basePopulation" name="neuron1Multi">
    <Parameter dimension="time" name="tau_p"/>
    <Parameter dimension="none" name="N"/>
    <Constant dimension="voltage" name="mVconst" symbol="mVconst" value="1mV"/>
    <Structure>
      <MultiInstantiate componentType="neuron1" number="N">
        <Assign property="v0" value="20*mVconst * index /  ( N-1 ) "/>
        <Assign property="tau" value="tau_p"/>
      </MultiInstantiate>
    </Structure>
  </ComponentType>
```

It is a bit tricky to compare the XML output between versions automatically because the order of elements and attributes can change (as in the examples above). Maybe it would be a good to use a tool like `xsltproc` and a script like `https://github.com/Appdynamics/xml-normalize` to have a "normalized" XML that you then can easily diff against another version.

Vigneswaran-Chandrasekaran:
Thank you very much for the comments and links :)
I made changes and now the output XML looks the same as the existing version. However, I checked only with the "working example" code in the documentation. I shall do some checking with other examples (also make some stylistic changes :smile:) and request for your kind review.

mstimberg:
Thanks for the work on this. I am now also getting the same output as in the previous version, but I wonder whether it is actually correct :worried: When I run it with [`jnml`](https://github.com/NeuroML/jNeuroML), I get a `NullPointerException`...

I wonder whether the approach to use `ComponentType` with `MultiInstantiate` can actually work like that, or whether we'll have to manually create one `Component` for each of the neurons. Maybe @pgleeson can shed some light on this? Here's a slightly simpler example: we have a LIF model with a constant input current, expressed as the equation `dv/dt = (v0 - v) / (10*ms)`. The `v0` should be a parameter of the neuron model which is different for each neuron in a population. The value for each neuron is given as a function of the index `i` of the neuron and the total number of neurons `n` as: `20*mV * i / (n-1)`.

Currently our NeuroML export expresses this like this:
```xml
<ComponentType extends="baseCell" name="neurongroup">
    <Property dimension="voltage" name="v0"/>
    <EventPort direction="out" name="spike"/>
    <Exposure dimension="voltage" name="v"/>
    <Dynamics>
      <StateVariable dimension="voltage" exposure="v" name="v"/>
      <TimeDerivative value="(v0 - v) / (10 * ms)" variable="v"/>
      <OnStart/>
      <OnCondition test="v .gt. (10 * mV)">
        <EventOut port="spike"/>
        <StateAssignment value="0*mV" variable="v"/>
      </OnCondition>
    </Dynamics>
  </ComponentType>
  <ComponentType extends="basePopulation" name="neurongroupMulti">
    <Parameter dimension="none" name="N"/>
    <Constant dimension="voltage" name="mVconst" symbol="mVconst" value="1mV"/>
    <Structure>
      <MultiInstantiate componentType="neurongroup" number="N">
        <Assign property="v0" value="20*mVconst * index /  ( n-1 )  "/>
      </MultiInstantiate>
    </Structure>
  </ComponentType>
  <network id="neurongroupMultiNet">
    <Component N="5" id="neurongroupMultipop" type="neurongroupMulti"/>
  </network>
```
When I run this with jNeuroML 0.8.5, I get a `NullPointerException` and:
```
SEVERE:  (ERROR) Accessed a value before it has been set? v0(Dimension[voltage m=1 l=2 t=-3 i=-1])
```

If you could give us a hint or point us to an example doing something similar @pgleeson we'd be very grateful :pray: 

Full LEMS file is in this [gist](https://gist.github.com/mstimberg/a990d274568dab0c56e283a11d144ebf)

pgleeson:
@mstimberg Happy to! Will be over the next day or two though...

mstimberg:
@pgleeson Thanks a lot. No rush of course!

pgleeson:
Main issue was that n and N were mixed up in one of the expressions. Changes here to make it run, and plot some traces: https://github.com/OpenSourceBrain/BrianShowcase/commit/3c05b8b3356360011d56f33b421c16a1add2178c#diff-85a8c40f0d8614e7bf68063a6642c8f954e7b38f61bdeffd396029dd23757987L25

Unfortunately there are still a few errors/warnings thrown related to the Property/Assign, which don't impact the running of the simulation, but might make debugging difficult.

mstimberg:
Ah, many thanks @pgleeson , I did not see that `n` issue :facepalm: It's actually a reference to an external constant `n` in the original Brian file, but this constant is identical to `N` so your replacement works. To make this work with arbitrary expressions we will have to add `<Constant .../>` definitions from our side.

@Vigneswaran-Chandrasekaran Could you take care of this? If an "initializer" refers to identifiers, these need to be included as `<Constant .../>` tags in the definition.

Vigneswaran-Chandrasekaran:
Yes sure. I think we have the `identifiers` information available in the `initializers_connectors` dictionary of `baseexport`, we shall directly use it with an additional check of that constant not used before with a different value

mstimberg:
Yes, sounds good!

mstimberg (approved):
Apologies for having taken so long, but all looks good to me nwo. I'll try to make a final pass over the code soon, so we can (finally) do a new release!

PR #43, opened 2020-10-26
update plot_morphology

mqcapelle:
Add option to plot compartments with colors corresponding to voltage value in that compartment

### DIFF
--- a/brian2tools/plotting/morphology.py
+++ b/brian2tools/plotting/morphology.py
@@ -2,50 +2,28 @@
 Module to plot Brian `~brian2.spatialneuron.morphology.Morphology` objects.
 '''
 import numpy as np
-from brian2.spatialneuron.spatialneuron import FlatMorphology
 
-from matplotlib.colors import colorConverter
+from matplotlib.colors import colorConverter, Normalize
 from matplotlib.patches import Circle, Polygon
 import matplotlib.pyplot as plt
 
+from brian2.spatialneuron.spatialneuron import FlatMorphology
 from brian2.units.stdunits import um
+from brian2.units.fundamentalunits import fail_for_dimension_mismatch
 from brian2.spatialneuron.morphology import Soma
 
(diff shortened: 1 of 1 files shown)


mstimberg:
Many thanks for the contribution! I will have a closer look soon, but it looks great at the first sight.

mstimberg:
Apologies for the time it took me to get back to this. This is a great contribution, I adapted it a bit to 1) make it no longer restricted to the membrane potential (e.g. you might want to plot currents, channel densitiy, distance from soma, ...) and 2) generalize the scaling a bit. For the latter, you can now either give two values as `value_norm` (mostly equivalent to your previous `voltage_min` and `voltage_max` solution, except that you are also allowed to leave it blank to take the minimum/maximum from the data), or give a colormap normalization function. Such functions simply map values to the range [0, 1], and matplotlib comes with e.g. `LogNorm` or `PowerNorm` which I guess could be useful in some situations.

@mqcapelle What do you think about the new syntax, does that work for you and your use cases (not sure you still have any, though ;) )?

mqcapelle:
>  I adapted it a bit to 1) make it no longer restricted to the membrane potential (e.g. you might want to plot currents, channel densitiy, distance from soma, ...) and 2) generalize the scaling a bit.
That is a good addition indeed. Cool that it now can be used for other properties as well.
Only thing is that, when users want to create a animation based on this function, `value_norm` must be explicitly set to maintain a consequent color scale.

>@mqcapelle What do you think about the new syntax
Yes looks great (:

mstimberg:
> Only thing is that, when users want to create a animation based on this function, `value_norm` must be explicitly set to maintain a consequent color scale.

Ah yes, good point.

I think I'll merge this as it is now, hopefully I'll find time soon to write up some documentation on this. I'll then mention this point. Maybe it would also be good to have an option for a colorbar.

mstimberg (approved):

PR #44, opened 2020-11-06
Model description polishing

mstimberg:
This makes a number of cosmetic/wording changes, but also adds two new major options to change the default behaviour:
* Monitors are no longer included by default in the description, since this information is not necessarily relevant. In order to include them, set `include_monitors` to `True` (the default value is `False`).
* Strictly speaking, initializers and `Synapses.connect` are not part of the model definitions and the order in which they are executed can matter. This is why they are stored in a list as part of the "run dictionary". In many cases, the initializations do not depend on the order of execution and it would be more natural to add the info to the definition of the respective group. This is now the default, to list them all at the end of network description, set `keep_initializer_order` to `True`. Ideally, we should analyze the code to figure out whether there are dependencies and either change the way of displaying it or raise an error/warning.

From my side this all looks good. @Vigneswaran-Chandrasekaran if you have time, could you have a look and see whether you find any issues (e.g. an example where it generates incorrect output)? Thanks!


mstimberg:
> From my side this all looks good.

Actually I'll still have to adapt the test suite, but apart from that it looks good :)

Vigneswaran-Chandrasekaran (approved):
Everything looks perfect from my side (Kindly review the documentation changes and mention any changes to be made)

mstimberg:
All looking good to me as well. There's a conflict with `master` I'll have to take care of but then I'll go ahead with the merge.

PR #47, opened 2020-11-27
Morphology plot improvements

mstimberg:
This adds a color bar for morphology plots with "values" (see #43) and also makes these plots work in 3D.

This might interest you @mqcapelle :)


mstimberg:
[Taking the liberty to merge my own PR]

PR #49, opened 2024-06-07
Added functionality of pandoc support

sparsh-989:
1. Used Pandoc inorder to provide support for file conversion to desired formats. Current available Format after this  
    P.R(Html, Pdf, Latex)

3. all(means all the above three format can be generated by using this functionality)

    This is what which need to be followed  in our script while adding the desired format
    Screenshot 2024-06-07 at 2 28 32 PM

### DIFF
--- a/brian2tools/mdexport/mdexporter.py
+++ b/brian2tools/mdexport/mdexporter.py
@@ -111,7 +111,7 @@ class MdExporter(BaseExporter):
                 raise Exception("Pandoc is not installed. Please install Pandoc and try again.")
             
             formats_extensions = {'latex':'.tex', 'html':'.html', 'pdf':'.pdf'}
-            if type(Additionalformat) == str:
+            if isinstance(Additionalformat, str):
                 if Additionalformat == "all":
                     formats = ['latex', 'html', 'pdf']
                 else:
--- a/brian2tools/mdexport/mdexporter.py
+++ b/brian2tools/mdexport/mdexporter.py
@@ -14,7 +14,7 @@ class MdExporter(BaseExporter):
     """
 
     def build(self, direct_call=True, debug=False, expander=None,
-              filename=None, Additionalformat=None, template_type=None):
+              filename=None, additonal_format=None, template_type=None):
         """
         Build the exporter
 
@@ -36,9 +36,10 @@ class MdExporter(BaseExporter):
             particular filename. When empty string '' is passed the user file
             name would be taken
 
-        Additionalformat : str and list of str
-                If user wants to have the output file in Additionalformat they
-                can specify them under this variable which is latex, html & pdf.    
+        additonal_format : str or list of str or all
+                If user wants to have the output file in additonal_format they
+                can specify them under this variable and the options are pdf, 
+                latex, html and all.  
 
         template_type : str   
             Based on your selected template, it will rendered otherwise
@@ -111,13 +112,13 @@ class MdExporter(BaseExporter):
                 raise Exception("Pandoc is not installed. Please install Pandoc and try again.")
             
             formats_extensions = {'latex':'.tex', 'html':'.html', 'pdf':'.pdf'}
-            if isinstance(Additionalformat, str):
-                if Additionalformat == "all":
+            if isinstance(additonal_format, str):
+                if additonal_format == "all":
                     formats = ['latex', 'html', 'pdf']
                 else:
-                    formats = [Additionalformat]
+                    formats = [additonal_format]
             else:
-                formats = Additionalformat if Additionalformat is not None else []
+                formats = additonal_format if additonal_format is not None else []
             for format_name in formats:
                 filename = self.filename + formats_extensions[format_name]
                 try:
--- a/brian2tools/mdexport/mdexporter.py
+++ b/brian2tools/mdexport/mdexporter.py
@@ -107,25 +107,25 @@ class MdExporter(BaseExporter):
             # Check if Pandoc is installed
             try:
                 subprocess.check_call(["pandoc", "--version"], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
-                formats_extensions = {'latex':'.tex', 'html':'.html', 'pdf':'.pdf'}
-                if type(Additionalformat) == str:
-                    if Additionalformat == "all":
-                        formats = ['latex', 'html', 'pdf']
-                    else:
-                        formats = [Additionalformat]
-                else:
-                    formats = Additionalformat if Additionalformat is not None else []
-                for format_name in formats:
-                    filename = self.filename + formats_extensions[format_name]
-                    subprocess.run(["pandoc", "--from", "markdown", "--to", format_name, "-o", filename, source_file])
-                    print("Conversion complete! Files generated:", filename) 
-                   
-            except subprocess.CalledProcessError:
+            except FileNotFoundError:
                 raise Exception("Pandoc is not installed. Please install Pandoc and try again.")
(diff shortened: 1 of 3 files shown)


mstimberg:
Many thanks for the PR @sparsh-989 – I will review it before our meeting.

brian2tools/mdexport/mdexporter.py
```diff
@@ -90,9 +95,26 @@ def build(self, direct_call=True, debug=False, expander=None,
...
             # start writing markdown text in file
-            md_file = open(self.filename + '.md', "w")
+            source_file = self.filename + ".md"
+            md_file = open(source_file, "w")
             md_file.write(self.md_text)
             md_file.close()
+            
+            # Check if Pandoc is installed
+            try:
+                subprocess.check_call(["pandoc"])
```
mstimberg, line 105: This does not work, it blocks, since `pandoc` tries to read from `stdin`. Instead, I'd suggest going back to `pandoc --version`, but suppressing the output.

brian2tools/mdexport/mdexporter.py
```diff
@@ -90,9 +95,26 @@ def build(self, direct_call=True, debug=False, expander=None,
...
+                formats_extensions = {'latex':'.tex', 'html':'.html', 'pdf':'.pdf'}
+                if format == "all":
+                    formats = ['latex', 'html', 'pdf']
+                else:
+                    formats = format
+                for format_name in formats:
```
mstimberg, line 111: This will raise an error for the default argument `None` – this needs to be handled. Also, it would be nice to accept a single string, e.g. `'latex'`, instead of forcing the user to specify `['latex']`.

brian2tools/mdexport/mdexporter.py
```diff
@@ -90,9 +95,26 @@ def build(self, direct_call=True, debug=False, expander=None,
...
+                    filename = self.filename + formats_extensions[format_name]
+                    subprocess.run(["pandoc", "--from", "markdown", "--to", format_name, "-o", filename, source_file])
+                    print("Conversion complete! Files generated:", filename)   
+            except subprocess.CalledProcessError:
```
mstimberg, line 115: This except catches only the missing installation of `pandoc`, any error that happens when running `pandoc` will be swallowed.

brian2tools/mdexport/mdexporter.py
```diff
@@ -34,6 +35,10 @@ def build(self, direct_call=True, debug=False, expander=None,
             If mentioned, the markdown text would be written in that
             particular filename. When empty string '' is passed the user file
             name would be taken
+
+
+        format : array
+                Array of string should be given as a input for this.
```
mstimberg, line 41: This needs to list the possible options. Also, it is not an `array`, but a `list` (or more general `sequence`/`iterable`). As I mention below, it would be good if it could accept a single string as well.

mstimberg (changes requested):
Hi. There are still a few things to fix, some of them we discussed during our last meeting. Most importantly, the current code does not run, since it blocks on the `pandoc` call.

brian2tools/mdexport/mdexporter.py
mstimberg:
```suggestion
                subprocess.check_call(["pandoc", "--version"], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
```

sparsh-989:
Additional format
user is not giving any input, let the .md file be there no error is expected, for non existing format error is expected
allow user to input single string rather then having a list for a single value
fix that panodc blocking
subprocess.run failing then have to handle error

sparsh-989:
details about format things for better readability

brian2tools/mdexport/mdexporter.py
```diff
@@ -14,7 +14,7 @@ class MdExporter(BaseExporter):
     """
 
     def build(self, direct_call=True, debug=False, expander=None,
-              filename=None, format=None, template_type=None):
+              filename=None, Additionalformat=None, template_type=None):
```
mstimberg, line 17: Please rename to `additional_formats` (see [PEP 8](https://peps.python.org/pep-0008/#function-and-variable-names))

brian2tools/mdexport/mdexporter.py
```diff
@@ -36,9 +36,13 @@ def build(self, direct_call=True, debug=False, expander=None,
             particular filename. When empty string '' is passed the user file
             name would be taken
 
+        Additionalformat : str and list of str
```
mstimberg, line 39: The type should be "str *or* list of str". And this should mention the "all" option, as well

brian2tools/mdexport/mdexporter.py
```diff
@@ -102,19 +106,26 @@ def build(self, direct_call=True, debug=False, expander=None,
             
             # Check if Pandoc is installed
             try:
-                subprocess.check_call(["pandoc"])
+                subprocess.check_call(["pandoc", "--version"], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
                 formats_extensions = {'latex':'.tex', 'html':'.html', 'pdf':'.pdf'}
-                if format == "all":
-                    formats = ['latex', 'html', 'pdf']
+                if type(Additionalformat) == str:
```
mstimberg, line 111: Normally, it is recommended to use `isinstance` instead of `type`, see e.g. https://docs.astral.sh/ruff/rules/type-comparison/

brian2tools/mdexport/mdexporter.py
```diff
@@ -102,19 +106,26 @@ def build(self, direct_call=True, debug=False, expander=None,
...
-                    formats = format
+                    formats = Additionalformat if Additionalformat is not None else []
                 for format_name in formats:
                     filename = self.filename + formats_extensions[format_name]
                     subprocess.run(["pandoc", "--from", "markdown", "--to", format_name, "-o", filename, source_file])
-                    print("Conversion complete! Files generated:", filename)   
+                    print("Conversion complete! Files generated:", filename) 
+                   
             except subprocess.CalledProcessError:
                 raise Exception("Pandoc is not installed. Please install Pandoc and try again.")
```
mstimberg, line 124: If `pandoc` is not installed, the error below will be triggered instead of this line. This exception here would currently only be raised if `pandoc` is installed, but `pandoc --version` fails nevertheless (not quite sure how this could happen).

brian2tools/mdexport/mdexporter.py
```diff
@@ -102,19 +106,26 @@ def build(self, direct_call=True, debug=False, expander=None,
...
-           
+            except FileNotFoundError as e:
+                raise Exception(f"Pandoc is not installed or not found in PATH: {e}")
+            except subprocess.SubprocessError as e:
```
mstimberg, line 127: This will never be triggered. Errors during the conversion are currently ignored (`run` by default runs with `check=False`).

mstimberg (changes requested):
Hi, I made a few comments for the most recent changes.

brian2tools/mdexport/mdexporter.py
sparsh-989: and available format

mstimberg (changes requested):
There's a typo in the word :-) Also, I'd suggest "additional_format**s**" to make it clear that the user can provide a list

mstimberg:
> There's a typo in the word :-) Also, I'd suggest "additional_format**s**" to make it clear that the user can provide a list

@sparsh-989 Could you please take care of this so that we can merge the PR?

mstimberg (approved):

PR #63, opened 2026-03-14
Enhance mdexport to annotate rand() and randn() using sympy

Utkarsha2811:
Problem: The Markdown exporter currently displays raw strings for initializations using rand() and randn() without providing mathematical context.

Solution: I used SymPy to analyze these expressions. The exporter now calculates and displays exact lower/upper bounds for rand() and the mean/variance for randn() in the exported Markdown descriptions.


Utkarsha2811:
Hey @mstimberg, I've merged master so the test suite will run. I also took the opportunity to do some minor refactoring and added tests while I was at it. Let me know what you think!

brian2tools/mdexport/expander.py
```diff
@@ -679,6 +679,73 @@ def expand_equations(self, equations):
...
+        tag) or an empty string when no annotation can be produced.
+        """
+        try:
+            import sympy
+            # Count occurrences in the raw string first: sympy deduplicates
+            # identical function calls (both rand() share the same placeholder
+            # argument), so expr.atoms() alone cannot distinguish one rand()
+            # from two rand() calls.
+            # Use a negative lookahead so that randn() is not counted as rand().
+            n_rand = len(re.findall(r'\brand(?!n)\s*\(', value_str))
```
mstimberg, line 705: I don't quite understand why this needs a negative lookahead. A simpler `\brand\s*\(` wouldn't match `randn()` either, would it?

brian2tools/mdexport/expander.py
```diff
@@ -679,6 +679,73 @@ def expand_equations(self, equations):
...
+            n_randn = len(re.findall(r'\brandn\s*\(', value_str))
+
+            expr = str_to_sympy(value_str)
+            funcs = expr.atoms(sympy.Function)
+
+            rand_funcs = [f for f in funcs if f.func.__name__ == 'rand']
+            randn_funcs = [f for f in funcs if f.func.__name__ == 'randn']
+
+            if n_rand == 1 and n_randn == 0 and len(rand_funcs) == 1:
```
mstimberg, line 714: Isn't the check `len(rand_funcs) == 1` redundant? What would be a situation where `n_rand == 1` but `len(rand_funcs) != 1`?

brian2tools/mdexport/expander.py
```diff
@@ -679,6 +679,73 @@ def expand_equations(self, equations):
...
+                r = rand_funcs[0]
+                min_val = sympy.simplify(expr.subs(r, 0))
+                max_val = sympy.simplify(expr.subs(r, 1))
+
+                var_tex = sympy.latex(str_to_sympy(variable_str), mode='plain')
+                min_tex = sympy.latex(min_val, mode='plain')
+                max_tex = sympy.latex(max_val, mode='plain')
+                annot_tex = fr"{var_tex} \in [{min_tex}, {max_tex}]"
+
+            elif n_randn == 1 and n_rand == 0 and len(randn_funcs) == 1:
```
mstimberg, line 724: Same question as for line 714

brian2tools/mdexport/expander.py
```diff
@@ -679,6 +679,73 @@ def expand_equations(self, equations):
...
+                # differentiate w.r.t. a Function application reliably.
+                mean_val = sympy.simplify(expr.subs(r, 0))
+                # For a linear expression f(randn()) = a*randn() + b the
+                # coefficient 'a' equals f(1) − f(0), i.e. the std deviation.
+                coeff = sympy.simplify(expr.subs(r, 1) - expr.subs(r, 0))
+                var_val = sympy.simplify(coeff ** 2)
+
+                mean_tex = sympy.latex(mean_val, mode='plain')
+                var_tex = sympy.latex(var_val, mode='plain')
+                annot_tex = fr"\mu={mean_tex},\ \sigma^2={var_tex}"
```
mstimberg, line 736: Both are of course possible, but I think I'd prefer having the std-dev instead of the variance, in particular because it has more natural units.

mstimberg (changes requested):
Hi again @Utkarsha2811 and another apology for the long time it took me to review. This looks mostly good to me, but I have a few minor comments.

PR #64, opened 2026-03-14
Feature: properly route multi-variable StateMonitors to subplots

### LINKED ISSUE #8: Allow plotting multiple variables of StateMonitor
At the moment, plotting a StateMonitor with multiple variables gives an error. It would be nice to plot multiple state variables to a figure with multiple axes (with shared x-axis so zooming in on one zooms on all). You could also imagine having a syntax that allows you to choose which axes are re-used and which are new, e.g. ``show_variables=[('v', 'vt'), ('I',)]`` would show v and vt on one axis, and I on another. It could automatically select the correct number of subplots to use to fill the figure.

The problem with this is that at the moment you can specify an axis to plot to, and this would have to allow plotting to multiple axes.

In the past I've written plotting routines for my own use where you can pass an ``ax`` and ``fig`` argument. If the plotting routine needs multiple axes, you have to pass a fig (or it creates a new one if fig is None). If the plotting routine only needs one axis, you can either pass an ax or fig (if you pass a fig, it will do ``subplot(111)`` on that fig). Maybe this scheme would work here?

Utkarsha2811:
Closes #8 

**Summary**

    1. brian_plot() now supports StateMonitor objects that record multiple variables. Instead of raising a TypeError, it creates stacked subplots with a shared time axis — one subplot per variable — and returns a list of Axes.
    2. var_name and var_unit accept per-variable dicts for custom labelling, e.g. var_name={'v': 'membrane potential', 'I': 'input current'}.
    3. The multi-variable logic is shared between the StateMonitor and StateMonitorView code paths via two private helpers (_resolve_var_kwds and _plot_state_variables), eliminating the previous code duplication.
    4. Users may still pass pre-created axes (one per variable); a clear TypeError is raised if the count doesn't match.

**Changes**

    1. brian2tools/plotting/base.py — extracted _resolve_var_kwds and _plot_state_variables helpers; simplified brian_plot's StateMonitor / StateMonitorView branches to two-line delegates; updated docstring.
    2. brian2tools/tests/test_plotting.py — added test_plot_multivar_monitors (auto-subplot, pre-created axes, wrong-length axes, StateMonitorView).
    3. docs_sphinx/user/plotting.rst — added "Multiple state variables" subsection with usage examples.


mstimberg (changes requested):
Thank you for the PR, I've left a few comments. Could you also add a test (only testing whether it works at all and returns the correct thing) and mention the feature in the documentation?

Utkarsha2811:
> Thanks for the changes, I won't be able to review this right away – please be patient 🙏

Hey @mstimberg, I’ve rebased the PR on top of master. Could you take a look when you get a chance? Tried to address issues mentioned previously.

mstimberg:
Hi @Utkarsha2811 Sorry for the very late reply, I was down with a cold recently and am very much behind with PR reviews… I'll hopefully find some time to review/test in detail tomorrow, but one thing I already saw: could you please include an image demonstrating this new feature in the documentation?

brian2tools/plotting/base.py
```diff
@@ -23,6 +23,71 @@
...
+
+    if axes is None:
+        fig, axes_arr = plt.subplots(n_vars, 1, sharex=True)
+    else:
+        axes_arr = np.asarray(axes).ravel()
+        if len(axes_arr) != n_vars:
+            raise TypeError(
+                "If multiple variables are recorded, 'axes' must be an "
+                "array-like of Axes with length %d (got %d)."
+                % (n_vars, len(axes_arr)))
```
mstimberg: Very minor detail, but I'd prefer an f-string over the old-school formatting used here.

brian2tools/plotting/base.py
```diff
@@ -23,6 +23,71 @@
...
+
+    ret_axes = []
+    for i, var_name in enumerate(record_variables):
```
mstimberg: Again a nitpicking detail, but I'd prefer the more Pythonic
```python
for ax, var_name in zip(axes_arr, record_variables):
```

brian2tools/tests/test_plotting.py
```diff
@@ -50,6 +50,46 @@ def test_plot_monitors():
...
+    plt.close()
+
+    # Wrong number of axes should raise TypeError
+    fig, bad_axes = plt.subplots(3, 1)
+    with pytest.raises(TypeError):
+        brian_plot(state_mon, axes=bad_axes)
+    plt.close()
+
+    # StateMonitorView of a multi-variable monitor
+    axes = brian_plot(state_mon[3])
```
mstimberg, line 85: I actually realized that an earlier comment of mine was incorrect – I thought e.g. `state_mon.v` would return a `StateMonitorView` (in the same way that for a `NeuronGroup` `group`, the access `group.v` would return a `VariableView`), but you are right that in `StateMonitor`, a "view" is used when you index into the group. I wonder whether it would make sense to change something on the Brian side so that you could call e.g. `brian_plot(state_mon.v)` (which currently does not work, since `state_mon.v` is a simple `Quantity` without information about the variable name). But this is outside the scope of this PR, of course.

mstimberg (changes requested):
Hi @Utkarsha2811, this looks all very good to me now. I made a few nitpicking comments below, but those are not essential. The only important missing thing is a picture in the documentation IMO.

Utkarsha2811:
Changes made as requested, @mstimberg. Thanks!

mstimberg (approved):
Apologies again for all the delays in reviewing this! Looks good to me now 🚀

PR #76, opened 2026-03-24
Fix baseexport: serialize hetero delays, capture resolved integrator, handle NetworkOperation

Sanchit2662:
## Summary
I was going through the baseexport collector and found three issues that were silently breaking things.                  
 
The `first` one was in collect_Synapses. The delay was only being serialized when it was a scalar value (single delay for all synapses). If you had per-synapse delays like S.delay = 'rand() * 5*ms', that condition would be false and the entire delay array would just get dropped from the export with no warning. The fix is literally removing that condition. obj.delay[:] works fine for both scalar and array cases, so there was no reason to guard it in the first place.           
  
The `second` one was in collect_NeuronGroup and collect_SpatialNeuron. When a user doesn't specify a method explicitly, Brian2 still picks one at runtime (like exact or exponential_euler) but we were just storing None in the export. There was even a TODO comment there acknowledging it. I fixed it by looking at the StateUpdater object's method_choice during the contained objects loop, where the runtime-resolved method is actually available. It gets stored as resolved_method so an importer can explicitly set the same integrator rather than relying on Brian2's auto-selection which might differ across versions.

The `third` one was NetworkOperation. If you had a @network_operation in your network, the exporter would just crash with a NotImplementedError. I added NetworkOperation to supported_objs so it no longer blows up, wrote a collect_NetworkOperation function that captures the scheduling metadata (name, dt, when, eorder), and added a logger warning so the user knows the function body itself isn't captured. It's honest about what it can and can't do rather than just crashing.

### DIFF
--- a/brian2tools/baseexport/collector.py
+++ b/brian2tools/baseexport/collector.py
@@ -49,8 +49,6 @@ def collect_NeuronGroup(group, run_namespace):
     # get user defined stateupdation method
     if isinstance(group.method_choice, str):
         neuron_dict['user_method'] = group.method_choice
-    # if not specified by user
-    # TODO collect from run time
     else:
         neuron_dict['user_method'] = None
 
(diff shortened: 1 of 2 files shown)


mstimberg:
Hi @Sanchit2662. Could you please merge latest `master` into this branch. The test suite should not be failing any more then.

